PR #117, opened 2022-08-12
#52 stop self reference in dependency wp

### LINKED ISSUE #52: Projects - Ensure Work Package Cannot Reference Self as Dependency in Endpoints
## Description

Make sure that the user can't add a work package's own WBS number as a dependency or the project that the work package is a part of as the work package's dependency. This functionality needs to be added to the:
- [ ] work package create endpoint
- [ ] work package edit endpoint

## Browser(s)

Chrome, Firefox, Safari

## Platform(s)

Desktop, iOS, Android

## Version(s)

main, production

## Observed Behavior

Currently, the application allows for a user to add a work package’s own WBS number to that work package’s list of dependencies. 

## Expected Behavior

- [ ] A user should not be able to set a work package to have itself as a dependency. 
- [ ] A user should not be able to set the project that a work package is within to be set as a dependency of the work package. 

Consider 1.1.1 which is a WP inside of 1.1.0. When adding dependencies to 1.1.1, don't allow the user to add 1.1.1 or 1.1.0 as dependencies.

## Proposed Solution

Look in the back-end where the work packages are created and edited, and ensure that WBS elements added to the dependencies list are not the same as the current work package or the project above. You should throw a 400 client error 
if an invalid dependency is attempted to be added. You will need to implement this functionality in 2 spots: 

- [ ] Work package create endpoint
- [ ] Work package edit endpoint

It is important to have this safeguard in all of these spots because users will have front-end form validation that errors early, before they click submit on the form. The back-end validation will also provide the stronger sense of security that the business logic will be correct. 

## Steps to Reproduce

Create a new work package with its anticipated WBS number in the list of dependencies. 

Edit a work package and add its WBSNumber to the list of dependencies. 

Create a new work package with the project WBSNumber also in the list of dependencies. 

Edit a work package and add the project it is in to the list of dependencies.

glacierA:
## Changes

Added checks in work package create and edit endpoints to stop requests where the dependencies include the package's own project or itself.

## Notes

Could possibly instead use an OR instead of two passes through the dependencies in the edit endpoint. Let me know if this is preferred.

Screen Shot 2022-08-12 at 4 07 22 PM

Closes #52

### DIFF
--- a/src/backend/src/controllers/work-packages.controllers.ts
+++ b/src/backend/src/controllers/work-packages.controllers.ts
@@ -105,8 +105,7 @@ export const createWorkPackage = async (req: Request, res: Response) => {
 
   if(dependencies.find((dep: { carNumber: any; projectNumber: any; workPackageNumber: any; }) =>
     dep.carNumber === carNumber
-    && dep.projectNumber === projectNumber
-    && dep.workPackageNumber === workPackageNumber) !== undefined) {
+    && dep.projectNumber === projectNumber && dep.workPackageNumber === workPackageNumber) !== undefined) {
     return res.status(400).json({ message: `A Work Package cannot have its own project as a dependency` });
   }
 
--- a/src/backend/src/controllers/work-packages.controllers.ts
+++ b/src/backend/src/controllers/work-packages.controllers.ts
@@ -103,8 +103,10 @@ export const createWorkPackage = async (req: Request, res: Response) => {
 
   if (workPackageNumber !== 0) throw new TypeError('Given WBS Number is not for a project.');
 
-  if(dependencies.find((dep: { carNumber: any; projectNumber: any; workPackageNumber: any; }) => dep.carNumber === carNumber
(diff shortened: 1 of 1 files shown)


src/backend/src/controllers/work-packages.controllers.ts
```diff
@@ -103,6 +103,13 @@ export const createWorkPackage = async (req: Request, res: Response) => {
 
   if (workPackageNumber !== 0) throw new TypeError('Given WBS Number is not for a project.');
 
+  if(dependencies.find((dep: { carNumber: any; projectNumber: any; workPackageNumber: any; }) =>
+    dep.carNumber === carNumber
```
jamescd18, line 107: I'm pretty sure I wrote a utility WBS Number equals function already

jamescd18 (changes requested):

src/backend/src/controllers/work-packages.controllers.ts
glacierA: Changed it to use existing.

anthonybernardi (approved):
nice!

PR #140, opened 2022-08-18
#138: iykyk

### LINKED ISSUE #138: Backend - fix user stuff
### Desired Changes

Adjust the user model

### Screenshots (as needed)

_No response_

anthonybernardi:
Adjust the user model

Closes #138

### DIFF
--- a/src/backend/src/utils/change-requests.utils.ts
+++ b/src/backend/src/utils/change-requests.utils.ts
@@ -6,6 +6,7 @@ import {
   StageGateChangeRequest,
   StandardChangeRequest
 } from 'shared';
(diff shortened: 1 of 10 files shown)


jamescd18 (changes requested):
I think there's more than needs scrubbing.

I think we need a user transformer to be able to limit which fields get sent to client. This one and all other instances of using users.

https://github.com/Northeastern-Electric-Racing/FinishLine/blob/685bd498a641e31a0a17f3f313a0c318e56efd5f/src/backend/src/utils/projects.utils.ts#L97

jamescd18 (approved):
LGTM

PR #141, opened 2022-08-18
#9: jwt

### LINKED ISSUE #9: Auth - Add Session JWT Validation
Utilize serverless approaches to issuing and validating signed JWT tokens for session validation. More definition and experimentation will be needed.

anthonybernardi:
## Changes

It works now! To test it, set `NODE_ENV="production"` in your backend env file and also you need the google auth id. For prod we need to make sure we set `NODE_ENV` and also `TOKEN_SECRET`

Closes #9

### DIFF
--- a/src/backend/src/utils/utils.ts
+++ b/src/backend/src/utils/utils.ts
@@ -39,3 +39,13 @@ export const authenticateToken = (req: Request, res: Response, next: any) => {
     next();
   });
 };
+
+export const requireJwtUnlessLogin = (fn: any) => {
+  return function (req: Request, res: Response, next: any) {
+    if (req.path === '/users/auth/:login') {
+      next();
+    } else {
+      fn(req, res, next);
+    }
+  };
+};
--- a/src/backend/src/utils/utils.ts
+++ b/src/backend/src/utils/utils.ts
@@ -2,6 +2,7 @@ import { Description_Bullet, WBS_Element, WBS_Element_Status } from '@prisma/cli
 import { DescriptionBullet, WbsElementStatus, WbsNumber } from 'shared';
 import jwt from 'jsonwebtoken';
 import { Request, Response } from 'express';
+import { TOKEN_SECRET } from '../..';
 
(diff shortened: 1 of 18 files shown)


src/frontend/src/apis/Users.api.ts
```diff
@@ -35,11 +35,19 @@ export const getSingleUser = (id: number) => {
...
-  );
+  return axios
+    .post<AuthenticatedUser>(
+      apiUrls.usersLogin(),
+      { id_token },
+      { transformResponse: (data) => authUserTransformer(JSON.parse(data)) }
+    )
+    .then((res) => {
+      const token = res.headers['authorization'];
+      localStorage.setItem('token', token);
```
jamescd18, line 46: I definitely don't think we should be storing tokens in local storage since that's the most easily accessed option

jamescd18 (changes requested):

src/backend/src/utils/utils.ts
```diff
@@ -20,3 +22,23 @@ export const convertStatus = (status: WBS_Element_Status): WbsElementStatus =>
     ACTIVE: WbsElementStatus.Active,
     COMPLETE: WbsElementStatus.Complete
   }[status]);
+
+export const generateAccessToken = (googleAuthId: string) => {
+  return jwt.sign({ googleAuthId }, process.env.TOKEN_SECRET as string, { expiresIn: '10h' });
+};
+
+export const authenticateToken = (req: Request, res: Response, next: any) => {
+  if (req.path === '/users/auth/:login' || process.env.NODE_ENV !== 'production') {
```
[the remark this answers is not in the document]
anthonybernardi: it is, the better question is why is this the url in the first place

src/frontend/src/apis/Users.api.ts
anthonybernardi: meh, tried to do httponly cookies but it doesn't work, get crazy cors errors

anthonybernardi:
bleh

PR #144, opened 2022-08-20
#67: Delete Risk Endpoint

### LINKED ISSUE #67: Projects - Endpoint for Deleting a Risk
### Description

Users may want to delete risks that are already resolved in order to remove them from the risk log and reduce clutter. A backend endpoint needs to be created to delete risks in the database.

### Prerequisites 

These tickets need to be completed before working on the current ticket:
- [x] Northeastern-Electric-Racing/PM-Dashboard-v2#698

### Proposed Solution

- [ ] Delete endpoint on the path string "/risks-delete"
- Accepted payload has: 
  - [ ] string `riskId` of the risk that is to be deleted
- Method will do the following:
  - [ ] Validate incoming data (all fields are present)
  - [ ] Validate that the risk is resolved. If the risk is not resolved, do not delete the risk. Return an error message.
  - [ ] If the risk is resolved
    - [ ] Set the `dateDeleted` field to the current time
    - [ ] Return a success message

anthonybernardi:
## Changes

added delete risk endpoint and also risk permissions

Closes #67


jamescd18 (dismissed):
LGTM

PR #147, opened 2022-08-26
#24-now-uses-proper-duration-util-function

### LINKED ISSUE #24: Projects - Fix Duration Calculation
## Current Behavior

The duration of projects is being calculated incorrectly as a simple sum of the durations of all work packages in the project. As many folks may remember from their interview, this is not the correct way to calculate the duration of a project.

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/092644cd1ebac0f9fcd70b458d8d5fa832409a08/src/backend/functions/projects.ts#L128

## Desired Behavior

The duration of a project is calculated by finding the latest `endDate` and earliest `startDate` among all the work packages in the project. Then perform `latestEndDate - earliestStartDate` and convert the result from days to weeks by dividing by 7. Ensure that the output is a whole number, no decimals.

glacierA:
## Changes

Now uses proper duration calculation function for project transformer return

## Notes

why was this function just sitting there in utils
Another grueling ticket.

## Screenshots (if applic
Screen Shot 2022-08-26 at 6 33 35 PM
able)

Closes #24

### DIFF
--- a/src/backend/src/utils/projects.utils.ts
+++ b/src/backend/src/utils/projects.utils.ts
@@ -6,7 +6,7 @@ import {
   DescriptionBullet,
   calculateEndDate,
   calculatePercentExpectedProgress,
-  calculateTimelineStatus
+  calculateTimelineStatus, calculateDuration
 } from 'shared';
 import { descBulletConverter, wbsNumOf } from './utils';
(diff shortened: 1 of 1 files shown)


jamescd18 (approved):
lmao LGTM!

Issue #98, opened 2022-03-31
Projects - Endpoints Should Use `res.status(40X).json({ message: 'x' })` Not `throw new TypeError()`


jamescd18:
## Description

Specifying errors on the 4xx level instead of 5xx would allow for project managers to independently troubleshoot any usage errors, omitting unnecessary follow-up with the software team.

## Background

Currently, the implementation of `throw new TypeError()` returns only 500 internal server error responses, which provides minimal feedback to users on what went wrong. Instead we should use express res to send a useful status and message

## Desired Behavior

4xx level errors should be returned.

## Proposed Solution

The `res.status(40X).json({message: 'blah blah blah'})` function should be implemented instead of `throw new TypeError()`, enabling error messages to be sent back to the front-end and be properly handled. Some reading material [here](https://www.tutorialspoint.com/session_initiation_protocol/sip_client_failure_responses.htm).

## Problem Areas

NOTE: These are outdated and from v2

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1062a1dc92e156db5141b1df4a57530d4b3446f5/src/backend/functions/work-packages.ts#L71

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1062a1dc92e156db5141b1df4a57530d4b3446f5/src/backend/functions/projects.ts#L106

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1062a1dc92e156db5141b1df4a57530d4b3446f5/src/backend/functions/users.ts#L25

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1062a1dc92e156db5141b1df4a57530d4b3446f5/src/backend/functions/work-packages-create.ts#L41

https://github.com/Northeastern-Electric-Racing/PM-Dashboard-v2/blob/1062a1dc92e156db5141b1df4a57530d4b3446f5/src/backend/functions/work-packages.ts#L71

PR #148, opened 2022-08-27
#98 - Implement try/catch responses in controllers

jonathan-chen10:
## Changes

There are two places `TypeError` shows up:

1. In functions like `createWorkPackage(req, res)` that have Express responses. In these cases the fix is straightforward: replace `TypeError` with the appropriate `res.status(40X)`. This is done in commit https://github.com/Northeastern-Electric-Racing/FinishLine/commit/dad227a34e7d0191fdc6273d8ab15cb3468a554c
1. In transformers like `userTransformer(user)` that return JSON data representing the user. This is either embedded within other transformers or sent to a controller that returns a `res`ponse. So our course of action is to **track all transformers** that rely at some point on `userTransformer` (the other problematics `teamTransformer` and `projectTransformer` also rely on `userTransformer`) and **try/catch `TypeError`s to return the appropriate response**.

## Notes

What if we treated all transformers as risky and not just the ones with `userTransformer`? This would make things more robust. All transformers except `authenticatedUserTransformer` rely on (and propagate errors from) `userTransformer` anyway. (`authenticatedUserTransformer` is only ever used in `logUserIn(req, res)`, so you can add the 400 error in there if you want).

We did not touch the transformers in `src/frontend/src/apis`, as they do not `throw new TypeError`.

## Test Cases

No tests yet. :( (Merging to working branch)

## To Do

Not sure how necessary these are. None of them are required for this fix to work.

- [ ] abstract out a helper function `tryUserTransformer(res, succResCode, failResCode)` (there was a lot of copy/paste here!)
- [ ] resolve `res.status` vs `return res.status`, which is currently inconsistent (find and replace `try { res` vs `try { return res`)
- [ ] as in the Notes, be extra cautious and add the try/catch to `authenticatedUserTransformer` too.

## Checklist

Please review the [contributor guide](https://github.com/Northeastern-Electric-Racing/FinishLine/blob/develop/docs/ContributorGuide.md) and reach out to your squad if anything is unclear.

- [ ] No linting errors
- [ ] No newline at end of file warnings
- [x] No merge conflicts
- [ ] All checks passing
- [x] Remove any not-applicable sections
- [x] No `package-lock.json` changes (unless dependencies have changed)

### DIFF
--- a/src/backend/src/controllers/change-requests.controllers.ts
+++ b/src/backend/src/controllers/change-requests.controllers.ts
@@ -9,7 +9,14 @@ import { Role } from '@prisma/client';
 
 export const getAllChangeRequests = async (req: Request, res: Response) => {
   const changeRequests = await prisma.change_Request.findMany(changeRequestRelationArgs);
-  return res.status(200).json(changeRequests.map(changeRequestTransformer));
+  try {
+    return res.status(200).json(changeRequests.map(changeRequestTransformer));
+  } catch (e) {
+    if (e instanceof TypeError) {
+      return res.status(400).json({ message: e.message });
+    }
+    throw e;
+  }
 };
 
 // Fetch the specific change request by its integer ID
@@ -22,7 +29,14 @@ export const getChangeRequestByID = async (req: Request, res: Response) => {
   if (requestedCR === null) {
     return res.status(404).json({ message: `change request with id ${crId} not found!` });
   }
-  return res.status(200).json(changeRequestTransformer(requestedCR));
+  try {
+    return res.status(200).json(changeRequestTransformer(requestedCR));
+  } catch (e) {
+    if (e instanceof TypeError) {
+      return res.status(400).json({ message: e.message });
+    }
+    throw e;
+  }
 };
 
 // handle reviewing of change requests
--- a/src/backend/src/controllers/projects.controllers.ts
+++ b/src/backend/src/controllers/projects.controllers.ts
@@ -20,7 +20,14 @@ import { descBulletConverter } from '../utils/utils';
 
 export const getAllProjects = async (_req: Request, res: Response) => {
   const projects = await prisma.project.findMany(manyRelationArgs);
-  res.status(200).json(projects.map(projectTransformer));
+  try {
+    res.status(200).json(projects.map(projectTransformer));
+  } catch (e) {
+    if (e instanceof TypeError) {
+      res.status(400).json({ message: e.message });
+    }
+    throw e;
+  }
 };
 
 export const getSingleProject = async (req: Request, res: Response) => {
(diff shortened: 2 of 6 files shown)


jamescd18:
I would agree with treating all back-end transformers as risky and try catching them. Front-end transformers are different tho so I think those should probably be left alone

jamescd18:
The front-end transformers could also perhaps be renamed to "hydration" functions or something like that, since I believe that's their primary purpose, to "rehydrate" string data from JSON which primarily is dates

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -9,7 +9,14 @@ import { Role } from '@prisma/client';
 
 export const getAllChangeRequests = async (req: Request, res: Response) => {
   const changeRequests = await prisma.change_Request.findMany(changeRequestRelationArgs);
-  return res.status(200).json(changeRequests.map(changeRequestTransformer));
+  try {
+    return res.status(200).json(changeRequests.map(changeRequestTransformer));
+  } catch (e) {
+    if (e instanceof TypeError) {
+      return res.status(400).json({ message: e.message });
```
jamescd18, line 16: I don't think 400 is necessarily the appropriate code

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -22,7 +29,14 @@ export const getChangeRequestByID = async (req: Request, res: Response) => {
   if (requestedCR === null) {
     return res.status(404).json({ message: `change request with id ${crId} not found!` });
   }
-  return res.status(200).json(changeRequestTransformer(requestedCR));
+  try {
+    return res.status(200).json(changeRequestTransformer(requestedCR));
+  } catch (e) {
+    if (e instanceof TypeError) {
```
jamescd18, line 35: I'm not so sure about the usage of TypeError. If you look up the docs it doesn't quite fit. If we think back to Java from fundies/OOD What we're missing is the different kinds of errors like illegal state, etc. Technically I think the ideal way is we subclass the Error object and make our own or something like that

src/backend/src/controllers/change-requests.controllers.ts
jamescd18: 400 indicates client error, but I don't super recall precisely what error states are possible

jamescd18 (commented):
there isn't anything inherently wrong with this ticket, but I'd say yes to reducing copy pasted code and I've also noted a couple key things with this sorta exception handling.

anthonybernardi:
Why do the transformers check for null at all? As can be seen below, it doesn't even let you pass something that could be null into the transformers. We don't need the try catches at all, you can just delete the type errors in the transformers. I don't know why they were there to begin with.

image

anthonybernardi (changes requested):
no try catches, just delete the throws in the transformers. Other stuff is good though

jonathan-chen10:
We decided to cancel this PR because we decided not to go the try/catch route. Instead, because the transformers can never accept `null` inputs, we will assume that Prisma will always pass good payloads to the transformers. We remove the `if (payload === null) throw new TypeError()` lines of the transformers in https://github.com/Northeastern-Electric-Racing/FinishLine/commit/cc42b46ef5df3870102070ba02586e8ad31d040c.

PR #150, opened 2022-08-27
#85: Proposed Solution schema changes

### LINKED ISSUE #85: Change Requests - N Proposed Solution Prisma Fix
## Description

Add in backend support to store an arbitrary amount of proposed solutions, with all appropriate information.

## Business Logic

We need to be able to support an arbitrary number of proposed solutions. To do this we need to add the following;
- [ ] A Change Request has a list of proposed solutions. `ProposedSolution[ ]`
- [ ] A `ProposedSolution` has the following:
  - [ ] A `Description` of the proposed solution, which is a `String`
  - [ ] A `TimelineImpact` for number of weeks impacted, which is an `int`
  - [ ] A `BudgetImpact` for dollar change needed, which is an `int`
  - [ ] A `Scope`, which is a `String`
 

## Proposed Solution

Adjust the Prisma backend to change the change request data type, to support the data forms from above.

anthonybernardi:
## Changes

added proposed solutions to the schema

## Notes

1. Not sure if this is the approach we want to go with for choosing the "approved" proposed solution
2. should we be taking out the other stuff from the normal scope CR? then we'll need a manual migration thing first but that's fine I guess
3. added formatting for the prisma file woohoo

Closes #85


src/backend/src/prisma/schema.prisma
```diff
@@ -61,30 +61,32 @@ model User {
...
-  createdRisks            Risk[]           @relation(name: "createdBy")
-  resolvedRisks           Risk[]           @relation(name: "resolvedBy")
-  deletedRisks            Risk[]           @relation(name: "deletedBy")
+  teams                   Team[]             @relation(name: "teamsAsMember")
+  teamAsLead              Team?              @relation(name: "teamAsLead")
+  createdRisks            Risk[]             @relation(name: "createdBy")
+  resolvedRisks           Risk[]             @relation(name: "resolvedBy")
+  deletedRisks            Risk[]             @relation(name: "deletedBy")
+  ProposedSolution        ProposedSolution[]
+  ProposedSolution        ProposedSolution[]
```
jamescd18, line 78: Two?

src/backend/src/prisma/schema.prisma
```diff
@@ -122,14 +124,30 @@ model Change_Request {
...
-  timelineImpact  Int
-  budgetImpact    Int
+  scopeCrId        Int                @id @default(autoincrement())
+  changeRequestId  Int                @unique
+  changeRequest    Change_Request     @relation(fields: [changeRequestId], references: [crId])
+  what             String
+  why              Scope_CR_Why[]
+  scopeImpact      String
+  timelineImpact   Int
+  budgetImpact     Int
```
jamescd18, line 134: Yes we should be removing these. Manual migration would be to convert all of the existing fields here being removed into a single proposed solution for each cr

src/backend/src/prisma/schema.prisma
```diff
@@ -122,14 +124,30 @@ model Change_Request {
...
+  ProposedSolution ProposedSolution[]
+}
+
+model ProposedSolution {
+  proposedSolutionId String   @id @default(uuid())
+  description        String
+  timelineImpact     Int
+  budgetImpact       Int
+  scope              String
```
jamescd18, line 143: scopeImpact

src/backend/src/prisma/schema.prisma
```diff
@@ -122,14 +124,30 @@ model Change_Request {
...
+  changeRequest      Scope_CR @relation(fields: [changeRequestId], references: [scopeCrId])
+  changeRequestId    Int
+  createdByUserId    Int
+  createdBy          User     @relation(name: "createdBy", fields: [createdByUserId], references: [userId])
+  dateCreated        Date     @default(now())
+  approvedBy         User?    @relation(name: "approvedBy", fields: [approvedByUserId], references: [userId])
+  approvedByUserId   Int?
```
jamescd18, line 150: I was thinking of having an optional field for "approvedSolution" on the CR that is the ID of the solution which is approved. What do you think? Otherwise I think maybe a Boolean on the proposed solutions for approved since person approving should already be recorded by who reviewed the cr.

jamescd18 (changes requested):
Dope to add that formatting. Commented all my thoughts and suggestions on data model changes

src/backend/src/prisma/schema.prisma
anthonybernardi: woops

src/backend/src/prisma/schema.prisma
anthonybernardi: we can't remove them right away. we need to have both for now, then run a manual script on prod, make sure that nothing is using these old ones, then we can finally delete them in a later ticket. (just my opinion but to me it's the safest)

src/backend/src/prisma/schema.prisma
anthonybernardi: is there ever a world where we have multiple proposed solutions? because doing it like this lets that world happen which is the only reason i think this could be better. If we are 100% against that forever then i agree it should be in the change request

src/backend/src/prisma/schema.prisma
anthonybernardi: true

src/backend/src/prisma/schema.prisma
jamescd18: sounds good if that's what you wanna do. me personally I would probably extract all the data first, turn it into a manual db script I can run on the new api, apply migration dropping all that data, and then run the manual script to put it back. but that's just me. I don't really care how it's done as long as it settles to where it needs to be. just use your best judgment 👍

src/backend/src/prisma/schema.prisma
anthonybernardi: after discussion, i still prefer the boolean on the proposed solutions. i just think it's nicer

src/backend/src/prisma/schema.prisma
anthonybernardi: hmmm ok

src/backend/src/prisma/schema.prisma
jamescd18: no, I don't think that would make sense. see slack thread: https://nu-electric-racing.slack.com/archives/C02U5TKHLER/p1661812958854149

src/backend/src/prisma/manual.ts
```diff
@@ -65,6 +64,32 @@ const activeUserMetrics = async () => {
   // select extract(week from "created") as wk, count(distinct "userId") as "# users", count(distinct "sessionId") as "# sessions" from "Session" group by wk order by wk;
 };
 
+/**
+ * migrate all Change Requests to use Proposed Solutions
+ */
+const migrateToProposedSolutions = async () => {
```
jamescd18, line 70: Does this cover approved CRs and stuff?

jamescd18: Aka does it cover marking which one is approved when the CR is approved and marking all as not approved if the CR is denied?

jamescd18 (changes requested):
I think just one thing missing on the manual script

jamescd18 (approved):
LGTM

PR #154, opened 2022-08-28
#18: added version number to settings page

### LINKED ISSUE #18: General - Add Version Numbering to the App for Deployments
## Description 

We should add some kind of version numbering to the application so that we can see what version users might have running, and so that we can see when the newest deployment has deployed.

## Background

These seem small but version number will likely be useful for much later in the future. Then we can just bump the version number whenever do deployments, which will likely be no more frequently than weekly given the nature of our team being students.

## Proposed Solution

- [ ] Research this feature and determine if this is a good idea. Will GitHub release features help achieve this?
- [ ] Implement with hard-coded string

## Mock

Untitled (Draft)-1

anthonybernardi:
## Changes

added version number to settings page

image

Closes #18

### DIFF
--- a/src/frontend/src/apis/misc.api.ts
+++ b/src/frontend/src/apis/misc.api.ts
@@ -1,0 +1,16 @@
+/*
+ * This file is part of NER's PM Dashboard and licensed under GNU AGPLv3.
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import axios from 'axios';
+import { apiUrls } from '../utils/Urls';
+
+/**
+ * Fetches the version number of the app.
+ */
+export const getVersionNumber = () => {
+  return axios.get(apiUrls.version(), {
+    transformResponse: (data) => JSON.parse(data)
+  });
+};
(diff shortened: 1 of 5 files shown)


jamescd18 (approved):
I was imagining having it at the very bottom of the left side bar so that it's visible anywhere in a screenshot of the app, but this works too if you prefer this

src/frontend/src/apis/misc.api.ts
```diff
@@ -0,0 +1,16 @@
...
+ * See the LICENSE file in the repository root folder for details.
+ */
+
+import axios from 'axios';
+import { apiUrls } from '../utils/Urls';
+
+/**
+ * Fetches the version number of the app.
+ */
+export const getVersionNumber = () => {
```
jamescd18, line 12: Just realized data is being pulled apart so this might be more correct to name it "get release info" or something like that. Or pull apart the data in this function

jamescd18 (changes requested):
I think slightly rename might be more accurate to reflect that more than just the version number is pulled from the api, but otherwise everything looks good!

PR #156, opened 2022-08-29
#131: Reset script is os agnostic

### LINKED ISSUE #131: Repo - Make the reset script OS agnostic
### Description

i wrote the `reset` script in package.json using mac commands so it doesn't work on windows. We need to make it so that it works on all OS's

### Background

-

### Proposed Solution

1. add this package as a dev dependency: https://github.com/isaacs/rimraf#cli
2. make the reset script use it
3. make sure it works

### Mocks

_No response_

anthonybernardi:
## Changes

added a package that lets me delete folders

Closes #131

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
     "build:frontend": "yarn workspace frontend build",
     "i": "yarn install && yarn prisma:generate",
     "istart": "yarn i && yarn start",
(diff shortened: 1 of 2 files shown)


jamescd18 (approved):
LGTM

PR #160, opened 2022-08-30
#135: Gantt Package Demo

### LINKED ISSUE #135: Work Packages - Investigate Alternate Gantt Chart Packages
## Spike Type

Technical

## Goal

Our goal is to switch over our existing Gantt chart package. At the successful conclusion of this epic, our work package Gantt charts will be swapped over to a selected new react package. No functionality will be lost, and Display dates will be changed from Sundays to Mondays. This package will be recorded as the new standard for our Gantt charts moving forwards

## Reason for Spike

As brought up by James and Anthony, we aren't too satisfied with the defacto bootstrap Gantt chart package we have been using. It isn't very intuitive and hard to customize for our needs. As such, we need to swap from it.

## Additional notes

We have some packages that Anthony has initially identified in React;
- [Frappe](https://github.com/frappe/gantt)
- [MaTeMaTuK](https://github.com/MaTeMaTuK/gantt-task-react)
- [ANovokmet](https://github.com/ANovokmet/svelte-gantt)
- [Bignumlock](https://github.com/bignumlock/react-gantt-elastic)

anthonybernardi:
## Changes

This package is really good: https://github.com/MaTeMaTuK/gantt-task-react

Here's a demo of it: https://matematuk.github.io/gantt-task-react/

It has critical path built in in a SUPER easy way.

## Notes

So I half assed my own demo (see below) because obviously this isn't the main ticket but there are a lot of good features on the demo site that we can include. For example, you can minimize the project which will collapse all the tasks (work packages) directly below it.

We can also make a custom component for the left side display thing which I think would probably be a good idea. Might be difficult but it can work.

There's also a lot of customization options for the visuals.

check out the gantt page icon woohoo
image

## To Do

might check out a different package. I think this is good enough but not totally sure.

Closes #135


anthonybernardi:
not gonna merge this, but this will definitely be used as an example for whoever does the gantt chart to start off with

PR #183, opened 2022-09-05
#87: modifed endpoint for approving change requests

### LINKED ISSUE #87: Change Requests - Adjust the Approve Change Request Endpoint
## Description

Update the Approve Change Request endpoint to reflect the new N Proposed Solutions Model

## Business Logic

Require that a Proposed Solution be selected by the user before a change request can be approved
- [ ] Enforce / block the ability to approve until the approver clicks on a proposed solution
- [ ] When this is done enable the ability to select the approve option
- [ ] When the button is pressed, add in the data corresponding to the selected proposed solution

Danwo0:
## Context

The existing endpoint for approving change request needs to be modified to reflect the N-Proposed Solutions model. 

## Changes

Added another check for approving to require that at least one proposed solution has been selected (`proposedSolution.approved`). Throws a 401 error if there is no solutions proposed, or none of the solutions are selected.
Approving process for other change requests are untouched.

## Notes

Change Request has been modified to ...

### Case 1: Success
Sending a POST request to the `/review` endpoint with:
```JSON
{
	"reviewerId": 3,
	"crId": 1,
	"reviewNotes": "test",
	"accepted": true
        "psId": "1"
}
```
The result should be:
`Status 200: OK`
```JSON
{
	"message": "Change request #1 successfully reviewed."
}
``` 
And update the `approved` field of the given proposed solution (psId: 1) to `true`.

### Case 2A: Fail: no psID provided
Sending a POST request without the psId field:
```JSON
{
	"reviewerId": 3,
	"crId": 1,
	"reviewNotes": "test",
	"accepted": true
}
```
The result should be:
`Status 400: Bad Request`
```JSON
{
	"message": "No proposed solution selected for scope change request"
}
``` 

### Case 2B: Fail: psID does not exist (or crId of the proposed solution does not match the crId being approved)
Sending a POST request without the psId field:
```JSON
{
	"reviewerId": 3,
	"crId": 1,
	"reviewNotes": "test",
	"accepted": true,
	"psId": "8"
}
```
The result should be:
`Status 400: Bad Request`
```JSON
{
	"message": "Proposed solution with id #8 not found for change request #1"
}
``` 

## Screenshots
With: 
image

### Case 1: Success
image

image

### Case 2A: Fail
image

image

### Case 2B: Fail
image

image

image

## Checklist

- [x] No linting errors
- [x] No newline at end of file warnings
- [x] No merge conflicts
- [x] All checks passing

Closes #87

### DIFF
--- a/src/backend/src/controllers/change-requests.controllers.ts
+++ b/src/backend/src/controllers/change-requests.controllers.ts
@@ -51,15 +51,17 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   if (reviewerId === foundCR.submitterId) return res.status(401).json({ message: 'Access Denied' });
 
   // make sure that a proposed solution is selected before approving
-  const foundPS = await prisma.proposed_Solution.findMany({ where: {changeRequestId: foundCR.crId }});
+  const foundPS = await prisma.proposed_Solution.findMany({
+    where: { changeRequestId: foundCR.crId }
+  });
   let selected = false;
   foundPS.forEach((proposedSolution) => {
-    if(proposedSolution.approved) {
+    if (proposedSolution.approved) {
       selected = true;
     }
   });
-  if (!selected) return res.status(401).json({message: 'No proposed solution selected'});
-  
+  if (!selected) return res.status(400).json({ message: 'No proposed solution selected' });
+
   // update change request
   const update = await prisma.change_Request.update({
     where: { crId },
--- a/src/backend/src/controllers/change-requests.controllers.ts
+++ b/src/backend/src/controllers/change-requests.controllers.ts
@@ -34,7 +34,7 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   }
 
   const { body } = req;
-  const { reviewerId, crId, reviewNotes, accepted, selectedPs } = body;
+  const { reviewerId, crId, reviewNotes, accepted, psId } = body;
 
   // verify that the user is allowed review change requests
   const reviewer = await prisma.user.findUnique({ where: { userId: reviewerId } });
@@ -50,15 +50,15 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   // verify that the user is not reviewing their own change request
   if (reviewerId === foundCR.submitterId) return res.status(401).json({ message: 'Access Denied' });
 
-  // make sure that a proposed solution is selected before approving
+  // if Scope CR, make sure that a proposed solution is selected before approving
   if (
     foundCR.type === 'ISSUE' ||
     foundCR.type === 'DEFINITION_CHANGE' ||
     foundCR.type === 'OTHER'
   ) {
-    if (!selectedPs) return res.status(400).json({ message: 'No proposed solution selected' });
+    if (!psId) return res.status(400).json({ message: 'No proposed solution selected' });
     await prisma.proposed_Solution.update({
-      where: { proposedSolutionId: selectedPs },
+      where: { proposedSolutionId: psId },
       data: {
         approved: true
       }
--- a/src/backend/src/controllers/change-requests.controllers.ts
+++ b/src/backend/src/controllers/change-requests.controllers.ts
@@ -51,12 +51,17 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   if (reviewerId === foundCR.submitterId) return res.status(401).json({ message: 'Access Denied' });
 
   // if Scope CR, make sure that a proposed solution is selected before approving
-  if (
-    foundCR.type === 'ISSUE' ||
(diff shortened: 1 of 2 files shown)


jamescd18 (approved):
I think it looks good to me, so long as we don't want to be marking all other proposed solutions as approved: false. This would mean accepted field on a proposed solution would basically only ever be undefined or true. I don't think that's a bad thing, just something to keep in mind

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -50,6 +50,21 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   // verify that the user is not reviewing their own change request
   if (reviewerId === foundCR.submitterId) return res.status(401).json({ message: 'Access Denied' });
 
+  // if Scope CR, make sure that a proposed solution is selected before approving
+  if (
+    foundCR.type === 'ISSUE' ||
```
anthonybernardi, line 55: a better way to check this is seeing `if (foundCR.scopeChangeRequest) { ... }`

however, to do this, you'll need to make sure prisma includes the scope cr on line 46

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -50,6 +50,21 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
...
+    foundCR.type === 'DEFINITION_CHANGE' ||
+    foundCR.type === 'OTHER'
+  ) {
+    if (!psId) return res.status(400).json({ message: 'No proposed solution selected' });
```
anthonybernardi, line 59: it's also worth checking if a proposed solution exists with that ID first because otherwise it will throw an error if it doesn't exist and you try to update it

anthonybernardi (changes requested):
Nice work! just some small changes

im also not sure if we can merge this yet until the actual part where the frontend starts sending the proposed solution id happens

anthonybernardi:
> I think it looks good to me, so long as we don't want to be marking all other proposed solutions as approved: false. This would mean accepted field on a proposed solution would basically only ever be undefined or true. I don't think that's a bad thing, just something to keep in mind

the default value in the schema is false for accepted

src/backend/src/controllers/change-requests.controllers.ts
Danwo0: Shoud I also check if the crId matches the one linked with the provided proposed solution?

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -34,7 +34,7 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   }
 
   const { body } = req;
-  const { reviewerId, crId, reviewNotes, accepted } = body;
+  const { reviewerId, crId, reviewNotes, accepted, psId } = body;
```
anthonybernardi, line 37: you have to update change-requests.routes.ts to include this as an optional parameter in the post

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -50,6 +50,27 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   // verify that the user is not reviewing their own change request
   if (reviewerId === foundCR.submitterId) return res.status(401).json({ message: 'Access Denied' });
```
anthonybernardi, line 52: to respond to your other comment, yeah we should be checking if the change request with crId exists in the first place (kinda a bug that that check doesnt exist already) and you can check that outside the `if foundscopecr` thing because we should check that always

anthonybernardi (changes requested):
very tiny changes. Great job!

src/backend/src/controllers/change-requests.controllers.ts
Danwo0: I believe that check is already being done around line 45; to clarify the question, I was wondering if I had to ensure that the crId that is in the proposoed solution matches the current crId that is provided/is being reviewed.

anthonybernardi: oof you're right i didn't see that. and yeah that's actually a very good point. you can just include that in the where part of the prisma query (so you have two constraints on the where now instead of one)

src/backend/src/routes/change-requests.routes.ts
```diff
@@ -20,6 +20,7 @@ changeRequestsRouter.post(
   body('crId').isInt({ min: 0 }).not().isString(),
   body('reviewNotes').isString(),
   body('accepted').isBoolean(),
+  body('psId').isString(),
```
anthonybernardi, line 23:
```suggestion
  body('psId').optional().isString().not().isEmpty(),
```

anthonybernardi (changes requested):
super tiny thing

PR #198, opened 2022-09-13
#72: Auto implement Activation CR

### LINKED ISSUE #72: Change Requests - Auto-Implement Approved Activation Change Requests
### Current Features

Currently there are no automated actions that serve to help implement a change request. Everything must be manually implemented after the change request has been approved.

Auto-implement an approved activation change request by:
- setting the associated work package to active
- if different, set the start date based on what is in the change request
- if different, set the project lead based on what is in the change request
- if different, set the project manager based on what is in the change request
- log implemented changes based on all the above changes, using the change request approver as the implementer

Also add a warning message to the CR page to explain approving this CR will trigger these actions.

### Screenshots (as needed)

_No response_

anthonybernardi:
## Changes

auto implements the stuff. Also fixed a tiny little bug (see if u find it)

this is the CR:
image

this is the wp right after activating:
(you can see the changes are logged correctly except for the date which is pretty weird...
image

## Todo
maybe figure out what's going on with the date?

Closes #72

### DIFF
--- a/src/backend/src/controllers/change-requests.controllers.ts
+++ b/src/backend/src/controllers/change-requests.controllers.ts
@@ -6,7 +6,7 @@ import {
   sendSlackChangeRequestNotification
 } from '../utils/change-requests.utils';
 import { validationResult } from 'express-validator';
-import { Role, WBS_Element_Status } from '@prisma/client';
+import { CR_Type, Role, WBS_Element_Status } from '@prisma/client';
 import { getUserFullName } from '../utils/users.utils';
 
 export const getAllChangeRequests = async (req: Request, res: Response) => {
@@ -67,7 +67,7 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   });
 
   // if it's an activation cr and being accepted, we can do some stuff to the associated work package
-  if (update.activationChangeRequest && accepted) {
+  if (update.type === CR_Type.ACTIVATION && update.activationChangeRequest && accepted) {
     const { activationChangeRequest: actCr, wbsElement } = update;
     const shouldUpdateProjLead = actCr.projectLeadId !== wbsElement.projectLeadId;
     const shouldUpdateProjManager = actCr.projectManagerId !== wbsElement.projectManagerId;
--- a/src/backend/src/controllers/change-requests.controllers.ts
+++ b/src/backend/src/controllers/change-requests.controllers.ts
@@ -6,7 +6,8 @@ import {
   sendSlackChangeRequestNotification
 } from '../utils/change-requests.utils';
 import { validationResult } from 'express-validator';
-import { Role } from '@prisma/client';
+import { Role, WBS_Element_Status } from '@prisma/client';
+import { getUserFullName } from '../utils/users.utils';
 
 export const getAllChangeRequests = async (req: Request, res: Response) => {
   const changeRequests = await prisma.change_Request.findMany(changeRequestRelationArgs);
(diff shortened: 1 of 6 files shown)


RChandler234 (commented):
This all looks great to me. Not sure wassup with the date, we should def fix that before approving the ticket tho. Does it fix itself if u refresh the page?

anthonybernardi:
fixing the date is a app-wide problem where we dont really convert timezones at all and it leads to these off by one day issues all the time. it could be that im just forgetting something stupid though

src/frontend/src/utils/Pipes.ts
```diff
@@ -28,6 +28,11 @@ export const wbsPipe = (wbsNum: WbsNumber) => {
   return `${wbsNum.carNumber}.${wbsNum.projectNumber}.${wbsNum.workPackageNumber}`;
 };
 
+/** Display WBS number as string but always the project number (1.2.3 -> 1.2.0) */
+export const wbsToProjectWbs = (wbsNum: WbsNumber) => {
```
jamescd18, line 32: Ew but okay

jamescd18: Ew cuz you're both destroying the object structure by converting to string and you're switching last digit to 0, so it's a two in one when we already have a function that does one of those two

jamescd18 (dismissed):
Ew one some things but I guess this should work and be okay

src/frontend/src/utils/Pipes.ts
```diff
@@ -28,6 +28,11 @@ export const wbsPipe = (wbsNum: WbsNumber) => {
...
+  return `${wbsNum.carNumber}.${wbsNum.projectNumber}.0`;
```
jamescd18, line 33:
```suggestion
  return wbsPipe({ ...wbsNum, workPackageNumber: 0 });
```

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -58,9 +62,71 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
       reviewNotes,
       accepted,
       dateReviewed: new Date()
-    }
+    },
+    include: { activationChangeRequest: true, wbsElement: { include: { workPackage: true } } }
   });
 
+  // if it's an activation cr and being accepted, we can do some stuff to the associated work package
+  if (update.activationChangeRequest && accepted) {
```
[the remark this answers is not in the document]
anthonybernardi: ill also check for type but it's also nice to have because it makes sure it's not undefined so i can pull it out and use it

src/backend/src/utils/users.utils.ts
```diff
@@ -33,3 +34,10 @@ export const userTransformer = (user: Prisma.UserGetPayload<null>): User => {
     role: user.role ?? undefined
   };
 };
+
+export const getUserFullName = async (userId: number | null) => {
+  if (!userId) return 'no one';
+  const user = await prisma.user.findUnique({ where: { userId } });
+  if (!user) return 'no one';
```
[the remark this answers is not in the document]
anthonybernardi: lol true, not sure of a better way

src/frontend/src/utils/Pipes.ts
anthonybernardi: tru

src/frontend/src/utils/Pipes.ts
anthonybernardi: nice

PR #209, opened 2022-09-19
#98 backend errors 4xx

### LINKED ISSUE #98: Projects - Endpoints Should Use `res.status(40X).json({ message: 'x' })` Not `throw new TypeError()`

sho-r1024:
## Changes

We added some tests to the work-packages.test.ts file, mainly tests with checking for a valid WBS Element ID and tests checking for null dependencies.

Other changes included: 

in wp controller : `throw TypeError` replaced with `return res.status(400)`
in `userTransformer`, `projectTransformer`, `teamTransformer`:  `TypeError` check for null input removed because TypeScript enforces that the input cannot be null

## To Do

Not sure if there's much else to do, but we'd like to double check before finalizing everything.

## Checklist

- [x] No linting errors
- [x] No newline at end of file warnings
- [x] No merge conflicts
- [x] All checks passing

Closes #98

### DIFF
--- a/src/backend/src/controllers/work-packages.controllers.ts
+++ b/src/backend/src/controllers/work-packages.controllers.ts
@@ -102,9 +102,9 @@ export const createWorkPackage = async (req: Request, res: Response) => {
   const { carNumber, projectNumber, workPackageNumber } = projectWbsNum;
 
   if (workPackageNumber !== 0) {
-    return res
-      .status(400)
-      .json({ message: `Given WBS Number ${projectWbsNum.toString()} is not for a project.` });
+    return res.status(400).json({
+      message: `Given WBS Number ${carNumber}.${projectNumber}.${workPackageNumber} is not for a project.`
+    });
   }
 
   if (dependencies.find((dep: any) => equalsWbsNumber(dep, projectWbsNum))) {
@@ -133,7 +133,9 @@ export const createWorkPackage = async (req: Request, res: Response) => {
   if (wbsElem === null) {
     return res
       .status(404)
-      .json({ message: `Could not find element with wbs number: ${projectWbsNum.toString()}` });
+      .json({
+        message: `Could not find element with wbs number: ${carNumber}.${projectNumber}.${workPackageNumber}`
+      });
   }
 
   const { project } = wbsElem;
@@ -165,12 +167,19 @@ export const createWorkPackage = async (req: Request, res: Response) => {
   const dependenciesIds: number[] = [];
   // populate dependenciesIds with the element ID's
   // and return error 400 if any elems are null
+
+  let dependenciesHasNulls = false;
   dependenciesWBSElems.forEach((elem) => {
     if (elem === null) {
-      return res.status(400).json({ message: 'One of the dependencies was not found.' });
+      dependenciesHasNulls = true;
+      return;
     }
     dependenciesIds.push(elem.wbsElementId);
   });
+
+  if (dependenciesHasNulls) {
+    return res.status(400).json({ message: 'One of the dependencies was not found.' });
+  }
 
   // add to the database
   await prisma.work_Package.create({
--- a/src/backend/src/controllers/work-packages.controllers.ts
+++ b/src/backend/src/controllers/work-packages.controllers.ts
@@ -162,11 +162,14 @@ export const createWorkPackage = async (req: Request, res: Response) => {
     })
   );
 
-  const dependenciesIds = dependenciesWBSElems.map((elem) => {
+  const dependenciesIds: number[] = [];
+  // populate dependenciesIds with the element ID's
+  // and return error 400 if any elems are null
+  dependenciesWBSElems.forEach((elem) => {
     if (elem === null) {
       return res.status(400).json({ message: 'One of the dependencies was not found.' });
     }
-    return elem.wbsElementId;
+    dependenciesIds.push(elem.wbsElementId);
   });
 
   // add to the database
--- a/src/backend/tests/work-packages.test.ts
+++ b/src/backend/tests/work-packages.test.ts
@@ -44,4 +44,11 @@ describe('Work Packages', () => {
 
     expect(res.statusCode).toBe(400);
   });
+
+  test('createWorkPackage fails if any elements in the dependencies are null', async () => {
+    const proj = { ...createWorkPackagePayload, dependencies: [null] };
+    const res = await request(app).post('/create').send(proj);
+
+    expect(res.statusCode).toBe(400);
+  });
 });
--- a/src/backend/tests/work-packages.test.ts
+++ b/src/backend/tests/work-packages.test.ts
@@ -1,0 +1,47 @@
+import request from 'supertest';
+import express from 'express';
+import workPackageRouter from '../src/routes/work-packages.routes';
+import { WBS_Element_Status } from '@prisma/client';
+import prisma from '../src/prisma/prisma';
+
+const app = express();
+app.use(express.json());
+app.use('/', workPackageRouter);
+
+// jest.mock('../src/utils/work-packages.utils');
+
+const createWorkPackagePayload = {
+  projectWbsNum: '1.2.0',
+  name: 'Pack your bags',
+  crId: 2,
+  userId: 1,
+  startDate: new Date('01/01/21'),
+  duration: 5,
+  dependencies: [
+    {
+      wbsElementId: 65,
+      dateCreated: new Date('12/25/20'),
+      carNumber: 1,
+      projectNumber: 1,
+      workPackageNumber: 1,
+      name: 'prereq',
+      status: WBS_Element_Status.COMPLETE
+    }
+  ],
+  expectedActivities: [{ descriptionId: 5, dateAdded: '01/05/21', detail: 'be active' }],
+  deliverables: [{ descriptionId: 3, dateAdded: '01/01/22', detail: 'did it work' }]
+};
+
+describe('Work Packages', () => {
+  beforeEach(() => {
+    prisma.work_Package.findMany = jest.fn();
+    prisma.work_Package.findUnique = jest.fn();
+  });
+
+  test('createWorkPackage fails if WBS number does not represent a project', async () => {
+    const proj = { ...createWorkPackagePayload, projectWbsNum: '1.2.1' };
+    const res = await request(app).post('/create').send(proj);
+
+    expect(res.statusCode).toBe(400);
+  });
+});
--- a/src/backend/tests/work-packages.test.ts
+++ b/src/backend/tests/work-packages.test.ts
@@ -1,26 +1,81 @@
 import request from 'supertest';
 import express from 'express';
 import workPackageRouter from '../src/routes/work-packages.routes';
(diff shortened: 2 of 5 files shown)


src/backend/src/controllers/work-packages.controllers.ts
```diff
@@ -101,10 +101,16 @@ export const createWorkPackage = async (req: Request, res: Response) => {
   // and what number work package this should be
   const { carNumber, projectNumber, workPackageNumber } = projectWbsNum;
 
-  if (workPackageNumber !== 0) throw new TypeError('Given WBS Number is not for a project.');
+  if (workPackageNumber !== 0) {
+    return res
+      .status(400)
+      .json({ message: `Given WBS Number ${projectWbsNum.toString()} is not for a project.` });
```
anthonybernardi, line 107: this will just print `Given WbS Number [object Object] is not for a project.` -- instead do something like `Given WBS Number ${carNumber}.${projectNumber} ...` blah blah blah etc yk

src/backend/src/controllers/work-packages.controllers.ts
```diff
@@ -156,12 +162,16 @@ export const createWorkPackage = async (req: Request, res: Response) => {
...
 
-  const dependenciesIds = dependenciesWBSElems.map((elem) => {
-    if (elem === null) throw new TypeError('One of the dependencies was not found.');
-    return elem.wbsElementId;
+  const dependenciesIds: number[] = [];
+  // populate dependenciesIds with the element ID's
+  // and return error 400 if any elems are null
+  dependenciesWBSElems.forEach((elem) => {
+    if (elem === null) {
+      return res.status(400).json({ message: 'One of the dependencies was not found.' });
```
anthonybernardi, line 170: this won't work because returning inside a forEach just ends the foreach and doesn't actually return from the outer function. set up a boolean outside like `dependenciesHasNulls` or something then after the forEach if it's true then you return the res 400

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,54 @@
+import request from 'supertest';
```
anthonybernardi, line 1: there is already a tests folder and a work-packages.test.ts file one directory above this

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,54 @@
...
+  deliverables: [{ descriptionId: 3, dateAdded: '01/01/22', detail: 'did it work' }]
+};
+
+describe('Work Packages', () => {
+  beforeEach(() => {
+    prisma.work_Package.findMany = jest.fn();
+    prisma.work_Package.findUnique = jest.fn();
+  });
+
+  test('createWorkPackage fails if WBS number does not represent a project', async () => {
```
anthonybernardi, line 41: you need to mock every prisma call that happens before the call to the endpoint finishes. For both of these, that means mocking `prisma.user.findUnique` and `prisma.change_Request.findUnique`. see the existing test files for how to do that

anthonybernardi (changes requested):

src/backend/src/controllers/work-packages.controllers.ts
sho-r1024: Why would we want to run the entire foreach loop if there's a null?

anthonybernardi: you're right, what I'm saying is that returning inside a foreach only ends the foreach (it's like `break`). the outer endpoint will keep running after that. You should end early if you see a null, but you also need a way to make sure the outer endpoint finishes too if you do that

src/backend/tests/work-packages.test.ts
sho-r1024: Where would I find the other tests folder? I can only seem to find one on my end

anthonybernardi: one directory above this. backend/tests

src/backend/tests/work-packages.test.ts
sho-r1024: Do we have to mock for prisma.work_Package.findUnique and prisma.work_Package.findAll? Also, would prisma.user.findUnique and prisma.change_Request.findUnique basically involve creating a user and creating a change request and then running jest.SpyOn() on both of them?

anthonybernardi: 1. in this case no you don't have to mock those because if the code is working as intended then the endpoint will fail before it even gets to those (returning the 400)
2. yeah pretty much. you can copy how that's done in the other files. it's kinda work intensive though so we should definitely in the future find a way to make it easier (probably by using pre made sample data that we can just import)

sho-r1024: @anthonybernardi Do I just test whether the two methods are called or is there more I have to do?

anthonybernardi: what two methods? the title of the test is what you're testing. For example, if you pass in 1 for work package number, it should fail right. That's already the case, we're just mocking any prisma calls along the way so that we aren't relying on prisma being in any certain state.

sho-r1024: `prisma.user.findUnique` and `prisma.change_Request.findUnique` are the two methods, how would I make the mock to test it? ig I'm just kinda confused about how the mocks in the other test files work?

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,154 @@
+import request from 'supertest';
+import express from 'express';
+import workPackageRouter from '../src/routes/work-packages.routes';
+import { CR_Type, Role, WBS_Element_Status } from '@prisma/client';
+import prisma from '../src/prisma/prisma';
+import { WbsElement } from 'shared';
```
anthonybernardi, line 6: unused, delete the line

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,154 @@
...
+const batman = {
+  userId: 1,
+  firstName: 'Bruce',
+  lastName: 'Wayne',
+  email: 'notbatman@gmail.com',
+  emailId: 'notbatman',
+  role: Role.APP_ADMIN
+};
+
+const nullReturn = () => {
```
anthonybernardi, line 23: unused, delete the function

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,154 @@
...
+import workPackageRouter from '../src/routes/work-packages.routes';
+import { CR_Type, Role, WBS_Element_Status } from '@prisma/client';
+import prisma from '../src/prisma/prisma';
+import { WbsElement } from 'shared';
+
+const app = express();
+app.use(express.json());
+app.use('/', workPackageRouter);
+
+jest.mock('../src/utils/work-packages.utils');
```
anthonybernardi, line 12: i don't think this is used anywhere

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,154 @@
...
+    jest.spyOn(prisma.wBS_Element, 'findUnique').mockResolvedValueOnce(someWBElement);
+    jest.spyOn(prisma.wBS_Element, 'findUnique').mockResolvedValueOnce(null);
+
+    const res = await request(app).post('/create').send(createWorkPackagePayload);
+
+    expect(res.statusCode).toBe(400);
+    expect(res.body.message).toBe('One of the dependencies was not found.');
+  });
+
+  test('findUniqueUser', async () => {
```
anthonybernardi, line 144: delete this entire test

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,154 @@
...
+      }
+    };
+    const res = await request(app).post('/create').send(proj);
+
+    expect(res.statusCode).toBe(400);
+    expect(res.body.message).toBe('Given WBS Number 1.2.2 is not for a project.');
+  });
+
+  test('createWorkPackage fails if any elements in the dependencies are null', async () => {
+    jest.spyOn(prisma.user, 'findUnique').mockResolvedValue({ ...batman, googleAuthId: 'b' });
```
anthonybernardi, line 133: no reason to change the google auth id right? you can just mock the resolved value to be batman

src/backend/tests/work-packages.test.ts
```diff
@@ -0,0 +1,154 @@
...
+describe('Work Packages', () => {
+  beforeEach(() => {
+    prisma.user.findUnique = jest.fn();
+    prisma.change_Request.findUnique = jest.fn();
+    prisma.work_Package.findMany = jest.fn();
+    prisma.work_Package.findUnique = jest.fn();
+  });
+
+  test('createWorkPackage fails if WBS number does not represent a project', async () => {
+    jest.spyOn(prisma.user, 'findUnique').mockResolvedValue({ ...batman, googleAuthId: 'b' });
```
anthonybernardi, line 116: same thing as my other comment

anthonybernardi (changes requested):
nice nice. After making my requested changes, see if the checks pass. this time the checks failed because the mocks weren't fully done so it tried to call prisma (but the github checks aren't hooked up to prisma so it fails). If it still fails (or the test check takes > 5 minutes) then that means you missed something to mock

src/backend/tests/work-packages.test.ts
sho-r1024: unfortunately TS throws an error saying "googleauthid is missing", not sure why

anthonybernardi (approved):
nice!

PR #217, opened 2022-09-21
#73 Auto-Implement Approved Stage Gate CR

### LINKED ISSUE #73: Change Requests - Auto-Implement Approved Stage Gate Change Requests
### Current Features

Currently there are no automated actions that serve to help implement a change request. Everything must be manually implemented after the change request has been approved.

Auto-implement an approved stage gate change request by:

- setting the associated work package to complete
- if not already, set the progress to 100%
- log implemented changes based on all the above changes, using the change request approver as the implementer

Also add a warning message to the CR page to explain approving this CR will trigger these actions.

### Screenshots (as needed)

_No response_

tigerhippo:
## Changes

1) Added CR_Type and WBS_Element_Status to the import statement 

2) Verified that the 'wbs element' variable exists, and if it was null, sent an HTTP message that the element was not found

3) Retrieved the 'progress' from the 'wbs element' variable

4) If the CR was accepted and the type of the CR is Stage Gate, updated the work package with the appropriate new information to be entered automatically for the logged changes

1) 
Screen Shot 2022-09-21 at 1 11 14 AM

2) 
Screen Shot 2022-09-21 at 1 11 34 AM

3) 
Screen Shot 2022-09-21 at 1 11 46 AM

4) 
Screen Shot 2022-09-21 at 1 12 03 AM

## To Do

It seems like nothing is showing up under 'Implemented Changes' even after I 'accept' the Stage Gate CR. There should be logged changes that appear automatically.

Screen Shot 2022-09-21 at 1 24 05 AM

## Checklist

- [X] No linting errors
- [X] No newline at end of file warnings
- [X] No merge conflicts
- [X] All checks passing
- [X] Screenshots of UI changes (if applicable)
- [X] PR is linked to the ticket

Closes # (issue #)

closes #73


src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -60,8 +60,57 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
...
+
+  if (wbsElement === null)
+    return res
+      .status(404)
+      .json({ message: `wbs element with id #${updated.wbsElementId} not found` });
+
+  const progress: number | undefined = wbsElement.workPackage?.progress;
+
+  if (updated.accepted && foundCR.type === CR_Type.STAGE_GATE) {
+    prisma.work_Package.update({
```
anthonybernardi, line 81: i spent like 10 minutes debugging before realizing it doesn't work because you didn't put await here lmao

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -60,8 +60,57 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
...
+        wbsElement: {
+          update: {
+            status: WBS_Element_Status.COMPLETE,
+            changes: {
+              createMany: {
+                data: [
+                  {
+                    changeRequestId: crId,
+                    implementerId: reviewerId,
+                    detail: 'Edited WBS element status from ' + wbsElement.status + ' to "COMPLETE"'
```
anthonybernardi, line 93: better way to do the string is `Edited WBS element status from "${wbsElement.status}" to "COMPLETE"`

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -60,8 +60,57 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
...
+                  },
+                  {
+                    changeRequestId: crId,
+                    implementerId: reviewerId,
+                    detail: 'Edited progress from "' + progress + ' to "100"'
```
anthonybernardi, line 98: better way to do the string is `Edited progress from "${progress}" to "100"`

anthonybernardi (changes requested):
ok so one more thing that I realized we forgot is that what if the progress is already at 100? we don't want to generate a change that says "change progress from 100 to 100" right. So you'll have to check if it's already 100 and if it isn't then you make the change and if it is then you don't make the change. you also have to do the same thing for status.

you can see how i did it in this pr https://github.com/Northeastern-Electric-Racing/FinishLine/pull/198

although actually there's a better way to do it. You can create the list just like i did but keep it all in the same query like you have it by passing the list in to the data on line 89

src/backend/src/controllers/change-requests.controllers.ts
anthonybernardi: tildes didn't show up woops, see above comment

anthonybernardi (approved):
awesome work!!!

PR #222, opened 2022-09-22
#132: updating issue templates

### LINKED ISSUE #132: Repo - Change Issue Templates
### Description
what even is an engineering task?

my real problem is that the background section doesn't make sense. If it's supposed to be "the background of this ticket" then that is covered by the description section. If it means "what do you need to know to be able to do this" then it shouldn't be called background

Pictured below:
Screen Shot 2022-08-17 at 1 35 05 PM

### Background

????

### Proposed Solution

change the template file pls

### Mocks

_No response_

anthonybernardi:
## Changes

Updated the issue templates a lot

closes #132

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature-request.yml
+++ b/.github/ISSUE_TEMPLATE/feature-request.yml
@@ -1,6 +1,6 @@
-name: New Feature
-description: For any issue related to creating a new feature
-title: "[Area] - Short Description"
+name: Feature Request
+description: Suggest a new feature for the project
+title: '[Area] - Short Description'
 labels: [new feature]
 body:
   - type: textarea
--- a/.github/ISSUE_TEMPLATE/task.yml
+++ b/.github/ISSUE_TEMPLATE/task.yml
@@ -1,6 +1,6 @@
-name: Engineering Tasks
-description: For issues related to engineering tasks
-title: "[Area] - Short Description"
+name: Task
+description: Create a task that a developer can complete.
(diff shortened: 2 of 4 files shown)


.github/ISSUE_TEMPLATE/task.yml
```diff
@@ -10,17 +10,17 @@ body:
     validations:
       required: true
   - type: textarea
-    id: background
+    id: acceptance-criteria
     attributes:
-      label: Background
-      description: What does someone need to know to work on this issue?
+      label: Acceptance Criteria
```
RChandler234, line 15: Yes.

.github/ISSUE_TEMPLATE/feature-request.yml
```diff
@@ -1,6 +1,6 @@
-name: New Feature
-description: For any issue related to creating a new feature
-title: "[Area] - Short Description"
+name: Feature Request
+description: Suggest a new feature for the project
```
RChandler234, line 2: Not sure when this would be used

leoRysing: The name feals funky since its basically a new feature, if we are adding something new from investigation

.github/ISSUE_TEMPLATE/task.yml
leoRysing: love love love acceptance criteria

leoRysing (approved):

Issue #169, opened 2022-09-01
Frontend - Rename TestSupport files


anthonybernardi:
### Desired Changes

when we were renaming the frontend files, we accidentally made some files in pascal case that shouldn't be since they're not react components. Rename the files in the TestSupport folder to be lowercase (and hyphenated between words in the case of change-requests.stub.ts).

Also rename the folder (and TestData folder) to be lowercase and hyphenated.

make sure tests are still passing after you do this (yarn test:frontend) and that everything compiles right and runs right and stuff

stevenl224:
renamed TestSupport folder and other files to lowercase and hyphenated format (ex: TestSupport --> test-support)

![Image](https://user-images.githubusercontent.com/108558702/191323397-19f06383-6850-4a5f-9cbd-cbea97750f90.png)

PR #234, opened 2022-09-27
#169: Renamed TestSupport files to kebab case

stevenl224:
Renamed TestSupport folder and files to kebab case from Pascal case.

Issues #169


anthonybernardi (approved):
awesome!

PR #235, opened 2022-09-27
#167: changed all stylesheet file names to kebab case

### LINKED ISSUE #167: Frontend - Rename stylesheets files
### Desired Changes

when we were renaming the frontend files, we accidentally made some files in pascal case that shouldn't be since they're not react components. Rename the files in the stylesheets folder to be lowercase (and hyphenated between words in the case of something like horizontal-list.module.ts).

make sure that all the imports of these files names are changed too (VSCode should hopefully do this for you)

make sure tests are still passing after you do this (yarn test:frontend) and that it runs right (`yarn:start`)

### Screenshots (as needed)

_No response_

ethankong150:
## Changes

changed all stylesheet file names to kebab case

Closes #167


jamescd18 (commented):
Folder names? Imports?

anthonybernardi (changes requested):
yeah to build off of james' review, folders should also be in the kebab case. also imports won't get changed by vscode but you can do a find all on vscode and search for like 'PascalCase.module' and replace all to be 'pascal-case.module'

Ari-S-123 (approved):
lgtm

anthonybernardi (changes requested):
running this locally it doesn't compile. seems like a few of the css and scss files still have capital letters but the imports have the correct lowercase. make sure after you change that everything looks like it's supposed to (you can compare with finishlinebyner.com) and if anythings unaligned or whatever then that means that something is broken

ethankong150:
the app seems to be compiling fine for me without any changes... not exactly sure what's happening.

Ari-S-123 (approved):
lgtm

anthonybernardi (approved):
nice!

PR #267, opened 2022-10-02
#224 - Added slackId parameter to User_Settings + updateUserSettings

### LINKED ISSUE #224: Backend - Add User Slack Ids
### Description

We want to be able to tie slack ids to users so that we can send them messages in the future from the app. The first step for this is adding that data to the database and adding an endpoint that allows us to edit it for any user.

https://www.workast.com/help/article/how-to-find-a-slack-user-id/

### Acceptance Criteria

1. The user settings model should have a `slackId` optional string field in the `schema.prisma`
2. change the `updateUserSettings` endpoint to include a slackId body parameter that will be incorporated into the prisma query and stuff to update the user's settings

### Proposed Solution

tips:
- make sure to run `yarn prisma:migrate` to generate the sql migration (see: https://nerdocs.atlassian.net/wiki/spaces/NER/pages/8749072/Software+Prisma+FAQ)
- change the validation for updateusersettings to be using express validator instead of in the function (see line `44` of users.controllers.ts)

sho-r1024:
## Changes

I did the two things requested in the ticket, namely adding the slackId parameter in User_Settings and then going into updateUserSettings and editing the endpoint within the method to take in a slackId and apply it to the User_Settings.

## To Do

- [ ] Any tests if necessary

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #224


.vscode/launch.json
```diff
@@ -0,0 +1,48 @@
+{
```
anthonybernardi, line 1: not sure what this file is but it shouldn't be part of this pr

src/backend/src/controllers/users.controllers.ts
```diff
@@ -41,14 +41,14 @@ export const getUserSettings = async (req: any, res: any) => {
 
 export const updateUserSettings = async (req: any, res: any) => {
   const userId: number = parseInt(req.params.userId);
-  if (!req.body || !req.body.defaultTheme) {
+  if (!req.body || !req.body.defaultTheme || !req.body.slackId) {
```
anthonybernardi, line 44: from the ticket "change the validation for updateusersettings to be using express validator instead of in the function (see line 44 of users.controllers.ts)"

src/backend/src/prisma/migrations/20221002215248_added_slack_id_to_user_settings/migration.sql
```diff
@@ -0,0 +1,11 @@
...
+  Warnings:
+
+  - You are about to drop the column `slackId` on the `User` table. All the data in the column will be lost.
+
+*/
+-- AlterTable
+ALTER TABLE "User" DROP COLUMN "slackId";
+
+-- AlterTable
+ALTER TABLE "User_Settings" ADD COLUMN     "slackId" TEXT;
```
anthonybernardi, line 11: this file is deleteing a column from user for some reason. i would delete this migration folder then reset your db then run migrate again

src/backend/src/controllers/users.controllers.ts
anthonybernardi: add it in the `user.routes.ts` as like `body('slackId').isString().not().isEmpty().optional({ nullable: true})` because that will allow it to either be undefined or null which is what we want. if you set something to undefined in prisma, it does nothing. if you set it to null, it turns it to null in the db. this way the frontend can send undefined if we want no change and null if we want to set it back to not having a slack id

anthonybernardi (changes requested):
nice start!

src/backend/src/prisma/migrations/20221002215248_added_slack_id_to_user_settings/migration.sql
sho-r1024: this was prob because I originally confused User_Settings with Users but I'll rerun migrate for good measure

src/backend/src/controllers/users.controllers.ts
sho-r1024: Should the error just be a json similar to how risks.controllers.ts does it (ex. ` return res.status(400).json({ errors: errors.array() });`)?

anthonybernardi: yeah the whole `  const errors = validationResult(req);
  if (!errors.isEmpty()) {
    return res.status(400).json({ errors: errors.array() });
  }` thing is at the start so that checks the express validator package stuff and if there were any errors as part of that it throws them

anthonybernardi: so just put that at the top of the endpoint. then you actually ahve to add the validation in the users.routes.ts file

src/backend/src/routes/users.routes.ts
```diff
@@ -12,7 +13,12 @@ const userRouter = express.Router();
 userRouter.get('/', getAllUsers);
 userRouter.get('/:userId', getSingleUser);
 userRouter.get('/:userId/settings', getUserSettings);
-userRouter.post('/:userId/settings', updateUserSettings);
+userRouter.post(
+  '/:userId/settings',
+  body('defaultTheme').isBoolean().not().isEmpty(),
```
jonathan-chen10, line 18: It looks like `defaultTheme` should be of type `ThemeName` (an enum), so this check should not be `isIn()` instead of `isBoolean()`. If we use `isIn()`, then we dont need `.not().isEmpty(),` either.

src/backend/src/routes/users.routes.ts
```diff
@@ -12,7 +13,12 @@ const userRouter = express.Router();
...
+  body('slackId').isString().not().isEmpty().optional({ nullable: true }),
```
jonathan-chen10, line 19: I think `notEmpty()` does the same as `not().isEmpty()` here.

jonathan-chen10 (changes requested):
Great job adding Express validation! I suggest a few minor tweaks to the validation.

src/backend/src/routes/users.routes.ts
anthonybernardi: i think notEmpty doesn't exist anymore? according to a quick google search idk

src/backend/src/controllers/users.controllers.ts
```diff
@@ -41,14 +42,16 @@ export const getUserSettings = async (req: any, res: any) => {
 
 export const updateUserSettings = async (req: any, res: any) => {
   const userId: number = parseInt(req.params.userId);
```
anthonybernardi, line 44: we should probably also check if `!userId` and error if so

src/backend/src/prisma/schema.prisma
```diff
@@ -97,10 +97,11 @@ model Session {
...
 model User_Settings {
-  id           String @id @default(uuid())
-  userId       Int    @unique
-  user         User   @relation(fields: [userId], references: [userId])
-  defaultTheme Theme  @default(DARK)
+  id           String  @id @default(uuid())
+  userId       Int     @unique
+  user         User    @relation(fields: [userId], references: [userId])
+  defaultTheme Theme   @default(DARK)
+  slackId      String?
```
anthonybernardi, line 104: we decided with the frontend ticket that instead of doing an optional slackId we're going to do a String with default empty string

to do this, delete your existing migration, change the schema.prisma, then run migrate again

src/backend/src/routes/users.routes.ts
```diff
@@ -12,7 +14,12 @@ const userRouter = express.Router();
 userRouter.get('/', getAllUsers);
 userRouter.get('/:userId', getSingleUser);
 userRouter.get('/:userId/settings', getUserSettings);
-userRouter.post('/:userId/settings', updateUserSettings);
+userRouter.post(
+  '/:userId/settings',
+  body('defaultTheme').isIn([Theme.DARK, Theme.LIGHT]),
+  body('slackId').isString().notEmpty().optional({ nullable: true }),
```
anthonybernardi, line 20: with the change to the schema this can just become .isString()

anthonybernardi (changes requested):
looks good! we're changing up a little bit about how we're storing the data but it should be a quick fix

src/backend/src/prisma/migrations/20221003023658_add_slack_id_param_to_user_settings/migration.sql
```diff
@@ -0,0 +1,2 @@
+-- AlterTable
+ALTER TABLE "User_Settings" ADD COLUMN     "slackId" TEXT;
```
anthonybernardi, line 2: as i said in my previous review, you need to delete this then run migrate again. right now this migration is not in line with what is in the prisma schema

src/backend/src/controllers/users.controllers.ts
```diff
@@ -41,14 +42,19 @@ export const getUserSettings = async (req: any, res: any) => {
...
-  if (!req.body || !req.body.defaultTheme) {
-    return res.status(404).json({ message: 'No settings found to update.' });
+  if (!userId) {
+    return res.status(404).json({ message: `could not find valid userId` });
+  }
```
anthonybernardi, line 47: lets also add a check here for getting hte user with the given userId from prisma, then if they dont exist that should be a 404 too

anthonybernardi (changes requested):
almost done, just gotta add one small thing and redo the migration pls

anthonybernardi (approved):
awesome!

PR #295, opened 2022-10-05
#294: Better Frontend Testing

### LINKED ISSUE #294: Frontend - Investigate Better Hook Testing Strategy
### Description

Right now testing the frontend is miserable because writing out the hooks for tests is painful. We need to try to find a way to fix that

### Acceptance Criteria

Propose a better alternative in a PR

### Proposed Solution

research stuff idk

### Mocks

_No response_

anthonybernardi:
## Changes

Look at how many lines I'm able to delete!!!!

basically we can use jest.spyOn which is only one line each and we can save the mock values in another file which makes this super compact and we only ever have to write the mock for each hook once.

## Notes

If we like this we can turn this into a lot of tickets

closes #294

### DIFF
--- a/src/frontend/src/tests/components/ActionButton.test.tsx
+++ b/src/frontend/src/tests/components/ActionButton.test.tsx
@@ -20,10 +20,6 @@ const renderComponent = () => {
 };
 
 describe('action button', () => {
-  it('renders without error', () => {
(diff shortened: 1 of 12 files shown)


Ari-S-123 (dismissed):
lgtm

Issue #142, opened 2022-08-19
MUI Migration


RChandler234:
## Description

Switch the frontend from using React-Bootstrap to the MUI component library. The goal is to maintain the look/functionality of the app as closely as possible while making the switch. 

For this ticket to be closed React-Bootstrap should be removed as a dependency and the changes should all be cleared with product.

Resources:
https://mui.com/

#### COMPONENTS/LAYOUT

- [x] #385
- [x] #386

#### PAGES

- [x] #388
- [x] #389
- [x] #390
- [x] #391
- [x] #392
- [x] #393
- [x] #394
- [x] #395
- [x] #396
- [x] #397
- [x] #398
- [x] #399
- [x] #400
- [x] #401
- [x] #402

### To Do as of 10/04/2022

- [x] Nav bar from above
- [x] #468
- [ ] #469
- [x] #470

PR #306, opened 2022-10-07
#142: MUI top bar

anthonybernardi:
## Changes

Screen Shot 2022-10-06 at 8 23 53 PM


RChandler234:
The Finishline by NER text is significantly bigger here than it is on the current version, it kinda draws focus from the welcome message in the center, lmk what you think. This is how it looks with the h5 typography variant instead for reference:
image

RChandler234 (approved):

PR #309, opened 2022-10-09
#204: Fixing Dates

### LINKED ISSUE #204: General - Find out why dates are inconsistent
### Problem

Dates are very inconsistent because we don't have a standardized day of dealing with them. For example, you might submit a change request on 9/15 and then it will get displayed as being created on 9/16 or something. Th exact cause of this is not known but I think it has to do with like say you submit a time on Eastern time zone, the backend when we give it a date assumes it is UTC, then when it sends it back it sends it as UTC and your browser converts it from UTC to Eastern. I'm not sure if this is exactly what's happening, but it's my best guess.

https://github.com/prisma/prisma/discussions/4153 this link shows that prisma's default time it creates is UTC, so this kinda backs up what I'm thinking where it seems to assume any time is in UTC even though we're passing it an EDT time.

### Steps to Reproduce

1. edit a work package start date or something to be a different day
2. see that it might show up wrong
3. this may or may not work depending on your time zone and your time of day

it happened to me in this pr here: https://github.com/Northeastern-Electric-Racing/FinishLine/pull/198

### Acceptance Criteria

What you should be doing in this ticket is seeing how to keep the exact time when you go from frontend input to backend output and back to frontend again. This will probably take a lot of console.logs or something and learning how javascript treats dates and timezones. If we can figure out the best process for keeping the same exact time throughout and exactly what is happening, we can fix it everywhere in another ticket. This ticket is just aiming to find out the WHY

anthonybernardi:
## Changes

1. makes the date pipe do the right conversions (very weird just trust me or read my documentation)
2. makes the work package setting the date not have weird not have weird UTC to EDT issues by making it be midday instead of 00:00. basically when you edit a wp to start on 10/08/2022 that gets set as 10/08/2022 at 00:00 am UTC in the db. Then, when this gets sent back to the frontend, it converts to EDT which is -4 hours so it bcomes 10/07/2022 at 8:00 pm which is weird so adding 12 hours gets rid of this problem with most or all time zones. idk it works for all the main ones

overall pretty happy with this because i think it covers the main issues. Probably still some more to it but whatever, should get rid of the biggest weird stuff

## Todo
- if you edit a work package but it goes from the same day to the same day it shouldn't log that as a change but it will because it will change the date from an actual time to 00:00 sometimes. might make this its own ticket because i think that is just a problem with seed data?

Closes #204

### DIFF
--- a/src/backend/src/controllers/work-packages.controllers.ts
+++ b/src/backend/src/controllers/work-packages.controllers.ts
@@ -131,11 +131,9 @@ export const createWorkPackage = async (req: Request, res: Response) => {
   });
 
   if (wbsElem === null) {
-    return res
-      .status(404)
-      .json({
-        message: `Could not find element with wbs number: ${carNumber}.${projectNumber}.${workPackageNumber}`
-      });
+    return res.status(404).json({
+      message: `Could not find element with wbs number: ${carNumber}.${projectNumber}.${workPackageNumber}`
+    });
   }
 
   const { project } = wbsElem;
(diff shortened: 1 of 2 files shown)


jamescd18 (approved):
lol okay lgtm

PR #319, opened 2022-10-10
#226: added and implemented helper print function

### LINKED ISSUE #226: Backend - Refactor Change string Creation
### Description

Changes have a detail that say ssomething like `Changed status from "0" to "100"`. We want this to be standardized across the app so that the formatting of this is consistent.

### Acceptance Criteria

1. there should be a new helper function in utils.ts called `buildChangeDetail` that takes in the parameters `thingChanged:string, oldValue: string, newValue: string`. this function should return a string in the format `Changed thingChanged from "oldValue" to "newValue"`
3. all instances of a change being created should use this helper function

### Proposed Solution

- changes should only be created in change-requests, work-packages, and projects controller files so look through those
- use the string templating \`text ${value} more text\` in the helper function

### Mocks

_No response_

stevenl224:
## Changes

Added and implemented a helper print function that printed the updated statuses of a change. Implementation of buildChangeDetail can be found in change-requests, work-packages, and projects controller messages. 

## Screenshots
- I had some issues with actually seeing the changes but the helper function logic appears correct and is implemented in the right places.
- 
image

Closes #226


RChandler234 (approved):

anthonybernardi (approved):
great work!

anthonybernardi (changes requested):
actually turns out there are more places that need to use this function, they just use the word "edited" and not "changed" so we didn't find them before:

`projects.utils.ts`: lines 241 and 365
`work-packages.utils.ts`: lines 108, 128, and 234

anthonybernardi (approved):
lgtm lets gooo

PR #324, opened 2022-10-12
322: Prisma Migrate DB on Deploy

### LINKED ISSUE #322: Devops - Prisma Migrate on Deploy
### Description

we can make `npx prisma migrate deploy` as part of our deployment pipeline which is good because this minimizes manual interactions with the database

### Acceptance Criteria

^

### Proposed Solution

add it to the dockerfile probably

### Mocks

_No response_

anthonybernardi:
## Changes

this should minimize manual db interaction which is great

Closes #322


jamescd18 (approved):
I think this looks good, and I guess makes sense. Though, it might be mildly iffy. I don't know how this would handle data-destructive migrations. Just be careful but I trust y'all will handle it fine.

PR #343, opened 2022-10-14
#342: dev login is better now

### LINKED ISSUE #342: Better Dev Login
### Description

make dev login not dependent on some weird array

anthonybernardi:
## Changes

When logging in as a dev, you can refresh as much as you want without being signed out!!! Also logging in is not based on some weird array and uses the actual db data which is cool.

also now themes work on local! you just have to refresh after setting it which i guess is fine (and it might be that way on prod too idk)

Closes #342


src/frontend/src/hooks/theme.hooks.ts
```diff
@@ -12,6 +12,7 @@ export const useProvideTheme = () => {
   const [theme, setTheme] = useState(themes[0]);
 
   const toggleTheme = (name: string) => {
+    console.log('got here ' + name);
```
RChandler234, line 15: remove

src/frontend/src/layouts/NavTopBar/NavUserMenu.tsx
```diff
@@ -31,23 +31,35 @@ const NavUserMenu: React.FC = () => {
...
+        ) : (
+          <button
+            className={'nav-link ' + styles.dropdownItems}
+            onClick={() => {
+              auth!.signout();
+              history.push(routes.LOGIN);
+            }}
+          >
+            Logout
+          </button>
```
RChandler234, line 61: what's this for?

src/frontend/src/pages/LoginPage/Login.tsx
```diff
@@ -36,11 +34,13 @@ const Login: React.FC<LoginProps> = ({ postLoginRedirect }) => {
...
 
-  const devFormSubmit = (e: any) => {
+  const devFormSubmit = async (e: any) => {
     e.preventDefault();
-    const user = exampleAllUsers.find((u) => u.role === devUserRole);
-    if (!user) throw new Error('user for dev not found from role: ' + devUserRole);
-    auth.devSignin(user!);
+    const authedUser = await auth.devSignin(devUserId);
+    if (authedUser.defaultTheme && authedUser.defaultTheme !== theme.name) {
+      console.log('toggle 2');
```
RChandler234, line 41: remove

src/frontend/src/pages/LoginPage/Login.tsx
```diff
@@ -21,7 +19,7 @@ interface LoginProps {
  * Page for unauthenticated users to do login.
  */
 const Login: React.FC<LoginProps> = ({ postLoginRedirect }) => {
-  const [devUserRole, setDevUserRole] = useState<string>(RoleEnum.APP_ADMIN as string);
+  const [devUserId, setDevUserId] = useState(1);
```
RChandler234, line 22: why is this set to 1 and not 0 as default? don't we have a userId of 1 in the database?

src/frontend/src/pages/LoginPage/LoginPage.tsx
```diff
@@ -46,13 +40,11 @@ const LoginPage: React.FC<LoginPageProps> = ({
           cookiePolicy={'single_host_origin'}
           isSignedIn={true}
         />
-        {process.env.NODE_ENV === 'development' ? (
-          <LoginDev devSetRole={devSetRole} devFormSubmit={devFormSubmit} />
-        ) : (
-          ''
-        )}
+        {process.env.NODE_ENV === 'development' ? <LoginDev devSetUser={devSetUser} devFormSubmit={devFormSubmit} /> : ''}
```
RChandler234, line 43: general question for the whole PR, but how secure is this using this process.env.NODE_ENV to check if we're on production/develop?

RChandler234 (changes requested):

src/frontend/src/layouts/NavTopBar/NavUserMenu.tsx
anthonybernardi: oh true thought it was prod only because of the google logout part but it works for dev

src/frontend/src/pages/LoginPage/Login.tsx
anthonybernardi: because then if you try to login without changing the dropdown at all it won't work

src/frontend/src/pages/LoginPage/LoginPage.tsx
anthonybernardi: uhhhhhh idk. the real security is on teh backend endpoint which definitely is secure so that's fine. if they try to dev signin on prod it just won't work so idk

anthonybernardi: the only not secure part if they can get around it on the frontend is that the dev login page calls getAllUsers which could be seen as a problem? but we don't restrict that anyways so idk

PR #357, opened 2022-10-19
#9: J W T

### LINKED ISSUE #9: Auth - Add Session JWT Validation

anthonybernardi:
## Changes

Gonna try to avoid using any keywords to avoid scrapers but basically it works. I was stuck before because it never worked on local but since I set up QA it turns out I was doing it mostly right and it was a localhost issue LOL

I was also able to set up better error messaging since I was already setting up the axios config file anyways and realized that was something I could do -- create a change request on prod with an invalid wbs (32.14.45 or something) and then do it on this localhost and see what the difference is. Or edit a project with an invalid crId (99999999) on prod and on local

Also changed our access denied error codes to 403 which is more accurate

## Test Cases

- manually tested on [QA](https://qa.finishlinebyner.com) -- if you need a higher role message me on Slack
- go to qa.api.finishlinebyner.com/users or /projects or anything and you'll see that it wont send you anything
- also manually tested with a 15 second expiration to see if that works and it does
- local dev still works on my computer but pls check on yours too just to get another set of eyes

## Future Improvements

some future improvements could be made...
- [ ] auto refreshing thetoken (as of now it doesnt log you out or refresh it, you just stop having access until you log back in lol). the reason this isn't a big deal rn tho is because every log in generates you a new jwt
- [ ] if it expires, log out the user
- [ ] need to actually check which user the jwt is so that we can line up the userId of the post request with the owner of the jwt token

Closes #9


src/backend/index.ts
```diff
@@ -11,17 +14,38 @@ import descriptionBulletsRouter from './src/routes/description-bullets.routes';
...
 const port = process.env.PORT || 3001;
 
+const allowedHeaders =
+  process.env.NODE_ENV === 'production'
+    ? 'Origin, X-Requested-With, Content-Type, Accept, Authorization, XMLHttpRequest'
+    : '*';
+
 const options: cors.CorsOptions = {
-  origin: '*',
+  origin: ['http://localhost:3000', 'https://finishlinebyner.com', 'https://magenta-mochi-275e56.netlify.app'],
```
ShaanHossain, line 23: is there a QA frontend url to add? I assume that's the last one, but adding a sub domain for it might be nice.

ShaanHossain: A common approach is api.host.com, qa.api.host.com, for API and QA API respective. host.com and qa.host.com for the frontend.

src/frontend/src/utils/axios.ts
```diff
@@ -0,0 +1,18 @@
+import { default as axiosStatic } from 'axios';
+
+const axios = axiosStatic.create({
+  withCredentials: process.env.NODE_ENV !== 'development' ? true : undefined
+});
+
+// This allows us to get good server errors
+// All express statuses must be: res.status(404).json({ message: "You are not authorized to do that." })
+axios.interceptors.response.use(
```
ShaanHossain, line 9: 🔥

src/backend/index.ts
anthonybernardi: good idea

src/backend/index.ts
```diff
@@ -11,17 +13,29 @@ import descriptionBulletsRouter from './src/routes/description-bullets.routes';
 const app = express();
 const port = process.env.PORT || 3001;
 
+// cors options
+const allowedHeaders = process.env.NODE_ENV === 'production' ? prodHeaders : '*';
 const options: cors.CorsOptions = {
-  origin: '*',
-  methods: 'GET,HEAD,PUT,PATCH,POST,DELETE',
```
ShaanHossain, line 16: Just wondering why we only use GETs and POSTs, feel free to put this in slack if you feel this discussion belongs there.

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -42,7 +42,7 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   const reviewer = await prisma.user.findUnique({ where: { userId: reviewerId } });
   if (!reviewer) return res.status(404).json({ message: `User with id #${reviewerId} not found` });
   if (reviewer.role === Role.GUEST || reviewer.role === Role.MEMBER)
-    return res.status(401).json({ message: 'Access Denied' });
```
ShaanHossain, line 45: Are these errors exposed directly to the user, or only the frontend programmers? 

If they are, Role X does not have access to do this might be more informative. It's my understanding that these messages are eaten by the frontend and would never surface to the user, so this is a moot point.

src/backend/src/controllers/change-requests.controllers.ts
```diff
@@ -51,7 +51,7 @@ export const reviewChangeRequest = async (req: Request, res: Response) => {
   if (foundCR.accepted) return res.status(400).json({ message: `This change request is already approved!` });
 
   // verify that the user is not reviewing their own change request
-  if (reviewerId === foundCR.submitterId) return res.status(401).json({ message: 'Access Denied' });
```
ShaanHossain, line 54: I also feel like it'd make more sense to turn this error message into a constant somewhere or create an error class we could return, so if we ever do want to change error messages, we can do so in one place.

src/backend/src/utils/utils.ts
```diff
@@ -24,3 +27,40 @@ export const convertStatus = (status: WBS_Element_Status): WbsElementStatus =>
 export const buildChangeDetail = (thingChanged: string, oldValue: string, newValue: string): string => {
   return `Changed ${thingChanged} from "${oldValue}" to "${newValue}"`;
 };
+
+const TOKEN_SECRET = process.env.TOKEN_SECRET || 'i<3security';
+
+// generate a jwt using the user's first and last name
+export const generateAccessToken = (user: { firstName: string; lastName: string }) => {
+  return jwt.sign(user, TOKEN_SECRET, { expiresIn: '12h' });
```
ShaanHossain, line 35: It's generally best practice to do 1 hour tokens, but I can't see any reason to pick 1, 12, or 24 hours. Just tossing it out there for consideration.

src/backend/src/utils/utils.ts
```diff
@@ -24,3 +27,40 @@ export const convertStatus = (status: WBS_Element_Status): WbsElementStatus =>
 export const buildChangeDetail = (thingChanged: string, oldValue: string, newValue: string): string => {
   return `Changed ${thingChanged} from "${oldValue}" to "${newValue}"`;
 };
+
+const TOKEN_SECRET = process.env.TOKEN_SECRET || 'i<3security';
```
ShaanHossain, line 31: I feel as though this should never have a default value. We'd want the service to crash if it isn't there. My general sentiment around default values is making sure that a default is truly a default, and not just a filler value. Why should 'i<3security' be our default token? And by embedding it directly in the code in a public repo that can be searched by bots, we're not doing ourselves any favors security wise.

For example, in java, we might use an Optional<Boolean>, so we can make it empty, true, or false vs. pretending false also represents non-existence.

anthonybernardi: something about express_jwt errors if we don't have the token secret so this is just default for development. we are absolutely never using this in any production ever

src/backend/src/controllers/change-requests.controllers.ts
anthonybernardi: yeah true i should make some tickets for that

src/backend/src/controllers/change-requests.controllers.ts
anthonybernardi: so i think that my changes to the axios file should make these error messages visible so yeah i should make some tickets to make these better

src/backend/index.ts
anthonybernardi: as of now we only have get and post requests so i'd figure i'd just limit it to that. we have a delete work package/project epic in the works so we'll probably add some delete endpoints for that. also i know that technically put and post have distinct meanings and stuff but i dont really see the point when they do the same thing ngl

RChandler234: what does taking them out of the methods here do?

RChandler234 (approved):

src/backend/index.ts
ShaanHossain: You will only be allowed to make POST or DELETE calls, none of the rest anymore.

ShaanHossain: "Taking them out" restricts someone from making the other types of HTTP requests.

src/backend/src/utils/utils.ts
ShaanHossain: Solid reasoning. My point still stands. You could omit the variable in prod by accident and never notice because it would build properly. 

However, I see your point even if it sacrifices production integrity.

src/backend/index.ts
ShaanHossain: Taking them out of methods will prevent someone from performing those types of requests on the backend. I.e. you will only be allowed to POST or DELETE

src/backend/src/utils/utils.ts
anthonybernardi: totally agree, just don't want to deal with the headache for new devs

src/backend/src/utils/utils.ts
anthonybernardi: 12 hours for now since we dont have auto refresh

PR #417, opened 2022-11-08
#414 - More Auth

### LINKED ISSUE #414: Auth - Send UserId with every query
### Description

we need to send the userid with every query probably as part of a auth header or something. this will let us do good things

another approach is to decode the jwt and attach userId there

anthonybernardi:
## Changes

this pr will actually prevent people from impersonating other users in requests which might be the last step in our auth journey. well, except we have to replace every instance of the userid in a request with getting it from the res object. I did this for a few risk endpoints as an example, it's very easy. We can turn that into easy tickets which is always nice.

it can be sorta weird to understand this since we have two totally different flows for login on dev and prod but here's the easiest way to explain them:

**Dev**: choose a user to login as >> backend is cool with it as long as the user exists >> save that userId in localstorage >> send that userId as a header with every request >> backend gets the userid from the header and puts it on the res.locals

**Prod**: login with google >> backend verifies that google was cool with it and if so generates the jwt >> jwt is a cookie so automatically gets sent back to the backend on every request >> backend decodes the jwt and says that this request is coming from whoever the jwt is for >> this userId is put on the res.locals

res.locals is the intended way to pass data between middleware. this means that we now know who a user is based off of their credentials, instead of just believing whoever they say they are in a post body. Right now, even though we have jwt, you can send a post with the userId being anyone you want and it doesn't check that that user lines up with whoever the jwt is issued for. that changes with this pr (after we convert all the endpoints to use the res.locals) which gets rid of impersonation and should be good enough auth for us for the foreseeable future.

## Test Cases

it works on dev

## To Do

- test on the QA environment
- remove all userId stuff from post bodies

Closes #414

### DIFF
--- a/src/backend/index.ts
+++ b/src/backend/index.ts
@@ -1,7 +1,7 @@
 import express from 'express';
 import cors from 'cors';
 import cookieParser from 'cookie-parser';
-import { prodHeaders, requireJwt } from './src/utils/utils';
+import { prodHeaders, requireJwtDev, requireJwtProd } from './src/utils/utils';
 import userRouter from './src/routes/users.routes';
 import projectRouter from './src/routes/projects.routes';
 import teamsRouter from './src/routes/teams.routes';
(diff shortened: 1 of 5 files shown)


jamescd18 (approved):
LGTM

RChandler234 (approved):

PR #427, opened 2022-11-10
#280: Changed to MUI Button. Uses NER attributes, but cannot use NERButton.

### LINKED ISSUE #280: Change Request Page Icon

nezamj03:
Closes #280 

@callmejim1226 @anthonybernardi @RChandler234

### DIFF
--- a/src/frontend/src/pages/ChangeRequestsPage/ChangeRequestsTable.tsx
+++ b/src/frontend/src/pages/ChangeRequestsPage/ChangeRequestsTable.tsx
@@ -117,28 +117,30 @@ const ChangeRequestsTable: React.FC = () => {
 
   return (
     <Container fluid>
-      <PageTitle
-        title={'Change Requests'}
-        previousPages={[]}
-        actionButton={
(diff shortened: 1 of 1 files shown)


src/frontend/src/pages/ChangeRequestsPage/ChangeRequestsTable.tsx
```diff
@@ -120,12 +121,22 @@ const ChangeRequestsTable: React.FC = () => {
         title={'Change Requests'}
```
RChandler234, line 121: Can you add a little space between the pageTitle and the table? You should be able to wrap the PageTitle in a div and use the style prop on that like the sx property on the mui button to set MarginBottom to like 10-20 or smthn, whatever looks best. Lmk if that doesn't make sense

RChandler234 (changes requested):

RChandler234 (approved):
🔥

callmejim1226 (approved):
LGTM! Great job!

PR #446, opened 2022-11-15
#445: refactoring some of the risks stuff

### LINKED ISSUE #445: Backend - Better Abstraction and Architecture
### Spike Type

Technical

### Goal

spread the backend out a little more to make it simpler

### Reason for Spike

there were some comments at the recent Leads meeting about the backend being confusing so I'm looking for ways to make it better

### Additional notes

_No response_

anthonybernardi:
## Changes

did some refactoring of the risk stuff and created a lot of error/response helper functions. I like this more because the services are actual functions with good inputs and outputs instead of just "request" and "response". I also think starting to standardize the response and error stuff is good (as suggested by Shaan).

I'm not sure if I'm doing the best possible way to do it with the error part though so could use some help there but it does work so idk

Closes #445

### DIFF
--- a/src/backend/src/services/risks.services.ts
+++ b/src/backend/src/services/risks.services.ts
@@ -1,0 +1,42 @@
+import { Role, User } from '@prisma/client';
+import { Risk } from 'shared';
+import prisma from '../prisma/prisma';
+import { throwAccessDeniedError, throwNotFoundError } from '../utils/response.utils';
+import { riskQueryArgs, riskTransformer } from '../utils/risks.utils';
+
+/**
+ * Gets all the risks for the given project
+ * @param projectId the project to get the risks for
+ * @returns a list of risks
+ * @throws if the given project doesn't exist
+ */
+export const getRisksForProject = async (projectId: number): Promise<Risk[]> => {
+  const requestedProject = await prisma.project.findUnique({ where: { projectId } });
+  if (!requestedProject) throwNotFoundError('Project', projectId);
+
+  const risks = await prisma.risk.findMany({ where: { projectId }, ...riskQueryArgs });
+
+  return risks.map(riskTransformer);
+};
+
+/**
+ * Creates a Risk in the database
+ * @param user the user creating the risk
+ * @param projectId the project to create the risk for
+ * @param detail the detail (description) of the risk
+ * @returns the id of the successfully created risk
+ * @throws if the user does not have access to create a risk
+ */
+export const createRisk = async (user: User, projectId: number, detail: string): Promise<string> => {
+  if (user.role === Role.GUEST) throwAccessDeniedError('Guests cannot create risks!');
+
+  const createdRisk = await prisma.risk.create({
+    data: {
+      project: { connect: { projectId } },
+      detail,
+      createdBy: { connect: { userId: user.userId } }
+    }
+  });
+
+  return createdRisk.id;
+};
--- a/src/backend/src/utils/response.utils.ts
+++ b/src/backend/src/utils/response.utils.ts
@@ -1,0 +1,31 @@
+import { Response } from 'express';
+
+export const sendErrorResponse = (res: Response, error: unknown) => {
+  if (error instanceof Error) {
+    res.status(error.cause as number).json({ message: error.message });
+  } else {
+    res.status(400).json({ message: 'Something went very wrong...' });
+  }
+};
+
+export const sendSuccessJsonResponse = (res: Response, obj: any) => {
+  res.status(200).json(obj);
+};
+
+export const sendSuccessMessageResponse = (res: Response, status: number, message: string) => {
+  res.status(status).json({ message });
+};
+
+export const throwError = (status: number, message: string): never => {
+  throw new Error(message, { cause: status });
+};
+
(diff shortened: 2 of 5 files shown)


src/backend/src/utils/response.utils.ts
```diff
@@ -0,0 +1,65 @@
+import { Response } from 'express';
+
```
ShaanHossain, line 2: [See this example](https://github.com/Northeastern-Electric-Racing/FinishLine/pull/447/files#r1023135726)

src/backend/src/services/risks.services.ts
```diff
@@ -0,0 +1,130 @@
+import { Role, User } from '@prisma/client';
```
ShaanHossain, line 1: Yea I think this service is perfect. The only question I have is whether we want to make it a class we export, so we can have a 1:1 naming scheme for the controller to service methods (at least some of them)

src/backend/tests/risks.test.ts
```diff
@@ -0,0 +1,22 @@
+import { getRisksForProject } from '../src/services/risks.services';
```
ShaanHossain, line 1: Sick tests

ShaanHossain:
My only qualm with this PR is that I'd like to see the entire risk.utils file refactored out into separate classes that handle those responsibilities, whether it's the service or a new file entirely. That file feels like it has too many responsibilities. 

The fact that it's named utils proves my point. We can't give it a simple name because it violates the single responsibility design principle.

src/backend/src/services/risks.services.ts
```diff
@@ -0,0 +1,130 @@
...
+
+/**
+ * Creates a Risk in the database
+ * @param user the user creating the risk
+ * @param projectId the project to create the risk for
+ * @param detail the detail (description) of the risk
+ * @returns the id of the successfully created risk
+ * @throws if the user does not have access to create a risk
+ */
+export const createRisk = async (user: User, projectId: number, detail: string): Promise<string> => {
```
callmejim1226, line 30: Really like how you commented on param, return, and throws description for every risk endpoint method. 
This way, it will be much easier for every ppl to understand our endpoint methods for risk, as it shows the description and data type for every parameter in every endpoint method.

src/backend/src/utils/response.utils.ts
```diff
@@ -0,0 +1,65 @@
...
+/**
```
callmejim1226, line 3: wow, another documentation for server response methods.. Impressive! XD

callmejim1226 (dismissed):
Liked this kind of refactoring! In my opinion, I think the key point for these changes was documentation for our risk endpoint methods and our general server response. Commenting out the description and data type for all parameters, returning data, and throwing msg for every method will allow ppl to understand our methods for risk endpoints much more easily! I bet our code for the backend will be much shortened in the future

src/backend/src/utils/errors.utils.ts
```diff
@@ -0,0 +1,53 @@
...
+
+export const errorHandler: ErrorRequestHandler = (error: unknown, req: Request, res: Response, next: NextFunction) => {
+  if (res.headersSent) {
+    return next(error);
+  }
+
+  if (error instanceof Exception) {
+    res.status(error.status).json({ message: error.message });
+  } else {
+    res.status(400).json({ message: 'Something went very wrong...' });
```
ShaanHossain, line 20: Why would we make this a 400 if something went very wrong? That sounds like a 500 - server side error to me?

src/backend/src/utils/response.utils.ts
```diff
@@ -0,0 +1,19 @@
...
+ * Send a success json response
+ * @param res the Express response object
+ * @param obj any object we want to send back (e.g., User, Project, Risk[], number)
+ */
+export const sendSuccessJsonResponse = (res: Response, obj: any) => {
```
ShaanHossain, line 8: I don't see the value in this method.

I feel like we'd want to control the type of success code, 201 for created, 204 for no-content on delete, 200 for Ok, most get responses

anthonybernardi: good point, i'll add the code.

the reason i want these methods though is more for the sendSuccessMessageResponse response, because i want to make it easy that we always send {message: 'message'}

src/backend/index.ts
```diff
@@ -48,6 +49,9 @@ app.use('/', (_req, res) => {
   res.json('Welcome to FinishLine');
 });
 
+// custom error handler middleware
+app.use(errorHandler);
```
ShaanHossain, line 53: perfect

src/backend/src/services/risks.services.ts
```diff
@@ -0,0 +1,133 @@
+import { Role, User } from '@prisma/client';
+import { Risk } from 'shared';
+import prisma from '../prisma/prisma';
+import { NotFoundException, AccessDeniedException, HttpException } from '../utils/errors.utils';
```
ShaanHossain, line 4: flawless service implementation. This is exactly what I'm talking about.

ShaanHossain (approved):
I'm comfortable with these refactors. If you merge yours in first, I'll update mine to make use of your error classes.

PR #447, opened 2022-11-15
#485 Refactor team controller and utils

ShaanHossain:
## Changes

Refactor teams controller and utils into
- Controller - manipulating the request and response
- Service - handling business logic, take in params from controller
- Transformer - transformer for converting the database response
- Args file - Store the prisma arguments

## Notes

- Added my vision of custom error handling


ShaanHossain:
I also want to hear some thoughts on making the service and controller into classes with methods we pull in. You mentioned not wanting it to look too much like Java, so I provided a small taste over here and you can let me know what you think.

src/backend/src/middleware/CustomErrorHandler.ts
```diff
@@ -0,0 +1,14 @@
+import { ErrorRequestHandler, Request, Response, NextFunction } from 'express';
```
ShaanHossain, line 1: This error handler lets us throw errors from the service, client, controller and sends it back to frontend accordingly.

src/backend/src/services/prisma.relation.args/team.args.ts
```diff
@@ -0,0 +1,13 @@
+import { Prisma } from '@prisma/client';
```
anthonybernardi, line 1: i dont hate this, just would maybe do dashes in the folder name and then `team.relation-args.ts` maybe?

anthonybernardi: also in the prisma folder or just the src folder

src/backend/src/transformers/teams.transformer.ts
```diff
@@ -1,19 +1,8 @@
 import { Prisma } from '@prisma/client';
```
anthonybernardi, line 1: definitely agree with having a transformers folder for just the transformers, forgot to do that in my pr

src/backend/src/services/teams.services.ts
```diff
@@ -0,0 +1,25 @@
+import { Team } from 'shared';
+import { NotFoundException } from '../exceptions/NotFoundException';
+import prisma from '../prisma/prisma';
+import { teamTransformer } from '../transformers/teams.transformer';
+import { teamRelationArgs } from './prisma.relation.args/team.args';
+
+export class TeamsService {
```
anthonybernardi, line 7: in this case I don't think the OOD aspect adds anything so I prefer functional

anthonybernardi (changes requested):
overall i think i prefer your error handling more, maybe with some tweaks, but prefer my functional aspect over the class based. also i agree with moving out the args and transformers from utils

src/backend/src/services/prisma.relation.args/team.args.ts
ShaanHossain: Done.

src/backend/src/services/teams.services.ts
ShaanHossain: We can't name the methods the same thing in the controller if we don't use the class. That's why I added it in the first place. What do you think the service methods should be named if not the same as their controller counterparts?

anthonybernardi: in my pr i had like `editRiskController` in the controller file and `editRisk` in the services file

src/backend/src/exceptions/CustomException.ts
```diff
@@ -0,0 +1,5 @@
+export abstract class CustomException extends Error {
+  constructor(public status: number, public message: string) {
```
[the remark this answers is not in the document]
anthonybernardi: i guess what I'm saying is that it doesn't get assigned to anything it just gets put in the constructor -- or is this just javascript class shorthand i dont know lol

src/backend/src/services/teams.services.ts
RChandler234: I like this class way, because when it's called its super obvious that a service is being called (which imo makes this concept of separating the business logic out into a service easier for someone else to figure out), as opposed to the functional ones looking like any other function. I also like that we can name it the same thing because I think it makes it a little easier for someone new/ names like editRiskController are a bit confusing since they're not controllers, they're just methods in a controller

RChandler234: @ShaanHossain @anthonybernardi

anthonybernardi: I actually think of it the other way around where adding the boilerplate of classes doesn't add any extra functionality. I think that if we do it like this then the controllers should also be class based though

src/backend/src/controllers/teams.controllers.ts
```diff
@@ -1,21 +1,18 @@
-import prisma from '../prisma/prisma';
 import { Request, Response } from 'express';
-import { teamRelationArgs, teamTransformer } from '../utils/teams.utils';
+import TeamsService from '../services/teams.services';
 
-export const getAllTeams = async (_req: Request, res: Response) => {
-  const teams = await prisma.team.findMany(teamRelationArgs);
-  return res.status(200).json(teams.map(teamTransformer));
-};
+export default class TeamController {
```
anthonybernardi, line 4: `TeamsController`

src/backend/src/services/teams.services.ts
```diff
@@ -0,0 +1,34 @@
+import { Team } from 'shared';
+import teamQueryArgs from '../prisma-query-args/team.query-args';
+import prisma from '../prisma/prisma';
+import teamsTransformer from '../transformers/teams.transformer';
+import { NotFoundException } from '../utils/errors.utils';
+
+export default class TeamsService {
+  /**
+   * Gets all of the teams across the side
```
anthonybernardi, line 9: across the side?

src/backend/tests/test-data/teams.test-data.ts
```diff
@@ -0,0 +1,26 @@
+import { Team as PrismaTeam } from '@prisma/client';
+import { Team as SharedTeam } from 'shared/src/types/team-types';
+import { flash } from './users.test-data';
+
+const teamId = 'id1';
+const teamName = 'pats';
+const slackId = '69';
+const description = 'best team no cap';
```
anthonybernardi, line 8: cap

anthonybernardi: (also i really like this organization to keep the prisma team and the shared team the same)

anthonybernardi (changes requested):
awesome, just some small stuff

PR #453, opened 2022-11-18
#165: Renaming API Files

### LINKED ISSUE #165: Frontend - Rename api files
### Desired Changes

when we were renaming the frontend files, we accidentally made some files in pascal case that shouldn't be since they're not react components. Rename the files in the apis folder to be lowercase (and hyphenated between words in the case of `change-requests.api.ts`).

make sure tests are still passing after you do this (`yarn test:frontend`)

### Screenshots (as needed)

_No response_

nezamj03:
closes #165


anthonybernardi (changes requested):
awesome!!!! one super small thing left is that the transformers folder name should be lowercase

callmejim1226 (approved):
Looks good to me! Great job!

anthonybernardi (approved):
awesome!

PR #456, opened 2022-11-22
#455: MUI Create CR Form

### LINKED ISSUE #455: Transition to MUI - Create Change Request Form
### Description

make it use mui and react hook forms

### Acceptance Criteria

mui

### Proposed Solution

mui

### Mocks

_No response_

anthonybernardi:
## Changes

fixed the create cr form and it fully works

## Notes

so the only hiccup is that I have to use the NativeSelect thing which does kinda look uglier but `<Select>` and `<TextField select>` both don't work with the react-hook-form useFieldArray and I cannot fix it for the life of me. (it also works with just normal `<select>` and `<input>` but then we get pure html woohoo so I think the NativeSelect is a goood enough middleground)

very annoying but I'm not sure it's worth spending more than the 30 minutes I've already spent trying to fix it. Also the NativeSelect won't change based on variation or styling or whatever so idrk what to do with that

## Screenshots

Screen Shot 2022-11-21 at 9 34 51 PM
Screen Shot 2022-11-21 at 9 34 30 PM

Closes #455

### DIFF
--- a/src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequest.tsx
+++ b/src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequest.tsx
@@ -39,10 +39,13 @@ const CreateChangeRequest: React.FC<CreateChangeRequestProps> = () => {
   } = useCreateProposeSolution();
   const [proposedSolutions, setProposedSolutions] = useState<ProposedSolution[]>([]);
 
-  const handleConfirm = async (data: FormInput) => {
-    if (auth.user?.userId === undefined) {
-      throw new Error('Cannot review change request without being logged in');
-    }
+  if (isLoading || cpsIsLoading || !auth.user) return <LoadingIndicator />;
+  if (isError) return <ErrorPage message={error?.message} />;
+  if (cpsIsError) return <ErrorPage message={cpsError?.message} />;
+
+  const { userId } = auth.user;
+
+  const handleConfirm = async (data: any) => {
     const crId = await mutateAsync({
       ...data,
       wbsNum: validateWBS(data.wbsNum),
@@ -53,7 +56,7 @@ const CreateChangeRequest: React.FC<CreateChangeRequestProps> = () => {
       const { description, timelineImpact, scopeImpact, budgetImpact } = ps;
       await cpsMutateAsync({
         crId,
-        submitterId: auth.user!.userId,
+        submitterId: userId,
         description,
         timelineImpact,
         scopeImpact,
@@ -64,9 +67,9 @@ const CreateChangeRequest: React.FC<CreateChangeRequestProps> = () => {
     history.push(routes.CHANGE_REQUESTS);
   };
 
-  if (isLoading || cpsIsLoading) return <LoadingIndicator />;
-  if (isError) return <ErrorPage message={error?.message} />;
-  if (cpsIsError) return <ErrorPage message={cpsError?.message} />;
+  const handleCancel = () => {
+    history.push(routes.CHANGE_REQUESTS);
+  };
 
   return (
     <CreateChangeRequestsView
@@ -75,6 +78,7 @@ const CreateChangeRequest: React.FC<CreateChangeRequestProps> = () => {
       onSubmit={handleConfirm}
       proposedSolutions={proposedSolutions}
       setProposedSolutions={setProposedSolutions}
+      handleCancel={handleCancel}
     />
   );
 };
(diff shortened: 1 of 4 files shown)


src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequest.tsx
```diff
@@ -39,10 +39,13 @@ const CreateChangeRequest: React.FC<CreateChangeRequestProps> = () => {
...
-    if (auth.user?.userId === undefined) {
-      throw new Error('Cannot review change request without being logged in');
-    }
+  if (isLoading || cpsIsLoading || !auth.user) return <LoadingIndicator />;
+  if (isError) return <ErrorPage message={error?.message} />;
+  if (cpsIsError) return <ErrorPage message={cpsError?.message} />;
+
+  const { userId } = auth.user;
+
+  const handleConfirm = async (data: any) => {
```
Ari-S-123, line 48: Can we use anything more specific than `any`?

src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequestView.tsx
```diff
@@ -8,27 +8,27 @@ import { Controller, useFieldArray, useForm } from 'react-hook-form';
...
 import Grid from '@mui/material/Grid';
 import CreateProposedSolutionsList from './CreateProposedSolutionsList';
+import ReactHookTextField from '../../components/ReactHookTextField';
+import { NativeSelect } from '@mui/material';
 
 interface CreateChangeRequestViewProps {
   wbsNum: string;
   crDesc: string;
-  onSubmit: (data: FormInput) => Promise<void>;
+  onSubmit: (data: any) => Promise<void>;
```
Ari-S-123, line 28: Here too.

Ari-S-123 (changes requested):

RChandler234 (commented):
If you try to submit with no Why's it just doesn't submit, but doesn't give you any error message/ prompting

z-roth:
picky ui thing, but i think it on larger screens it would make more sense to put wbs number,  type, and what on the same row, and whys in the rows below it.

src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequestView.tsx
```diff
@@ -81,166 +68,141 @@ const CreateChangeRequestsView: React.FC<CreateChangeRequestViewProps> = ({
...
+            <Box>
+              <Typography variant="caption">Type</Typography>
+            </Box>
+            <Controller
+              name="type"
+              control={control}
+              rules={{ required: true }}
+              render={({ field, fieldState }) => (
+                <FormControl>
+                  <NativeSelect {...field} variant="outlined" error={!!fieldState.error}>
```
z-roth, line 125: saw someone online did an implementation by creating a custom controller component, worth a look if you haven't seen it 

https://stackoverflow.com/questions/63236951/how-to-use-mui-select-with-react-hook-form

z-roth: might save us some time in future implementations as well

anthonybernardi: i already made some for other tickets like ReactHookTextField and ReactHookEditableList

this is a fringe case caused by the array of objects and dropdowns that we're not going to get around easily

src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequestView.tsx
```diff
@@ -81,166 +69,143 @@ const CreateChangeRequestsView: React.FC<CreateChangeRequestViewProps> = ({
...
+          <Grid item xs={12} md={9}>
+            <Box>
+              <Typography variant="caption">Type</Typography>
+            </Box>
+            <Controller
+              name="type"
+              control={control}
+              rules={{ required: true }}
+              render={({ field, fieldState }) => (
+                <FormControl>
```
RChandler234, line 125: can we really not do this one with the MUI select? The other case in this form makes sense, but this one seems pretty simple/ similar to other stuff we've done

RChandler234 (changes requested):

src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequestView.tsx
anthonybernardi: we defintely can i just wanted it to be consistent on the page

PR #467, opened 2022-12-03
#64: Fix tooltips extending past screen

### LINKED ISSUE #64: Change Requests - Changes Tooltips Extend Past Window Width
### Browser(s)

Chrome, Firefox, Safari

### Platform(s)

Desktop

### Version(s)

main, production

### Observed Behavior

The tooltip that pops up next to an implemented change bullet can extend beyond the window width and produce a horizontal scroll bar.

### Expected Behavior

Ideally the tooltip is constrained within the window width, with no horizontal page scrolling. 

Consider whether alternative solutions such as moving the tooltip to the top/bottom of the bullet point instead of to the right side would be preferred.

### Steps to Reproduce

1. Navigate to a Projects, Work Package, or Change Requests details page that has a bulleted list of Changes (Project, WP page) or Implemented Changes (CR page).

    Examples in main:
     - Project 1.1.0 -> Changes
     - Work Package 1.1.2 -> Changes
     - Change Request 6 -> Implemented Changes

    Examples in prod:
     - Project 0.1.0 -> Changes
     - Work Package 0.1.2 -> Changes
     - Change Request 7 -> Implemented Changes

2. Shrink the browser window to the minimum possible width.
3. In the (Implemented) Changes section, hover over the longest bullet point to reveal the tooltip. If the change text is long enough, the tooltip will extend beyond the window width.

### Screenshots (as needed)

from 1.1.0 in main (local) version:  
Before Hover
Screen Shot 2022-05-28 at 11 06 34 PM
On Hover
Screen Shot 2022-05-28 at 11 05 56 PM

zachnorman02:
## Changes

Before, the tooltips were to the right and would cut off past the screen if they were too long. I was playing around with either just keeping them always on the top, or changing them from right to top depending on the size of the screen

## Notes

The tooltips were working before, but I'm having an issue now where the tooltips are just defaulting to how they're implemented currently and not updating. I even tested changing the text of the tooltips to see if something was just weird with the tooltip position, but the text won't update either. The React Developer tools also shows the OverlayTrigger's placement prop being set to 'right', even when I remove what I currently have committed and just replace all of it with 'top' (so no mention of 'right' anywhere in the prop value). I'm trying to figure out what the issue is but I'm getting stuck, especially since this exact code was working a couple weeks ago.

## Screenshots

*Need to insert once I can figure out why the tooltips aren't updating*

## To Do

- [ ] Make sure tooltips display properly and the way that's expected

## Checklist

- [x] No merge conflicts
- [ ] All checks passing

Closes #64

### DIFF
--- a/src/frontend/src/pages/ChangeRequestDetailPage/ImplementedChangesList.tsx
+++ b/src/frontend/src/pages/ChangeRequestDetailPage/ImplementedChangesList.tsx
@@ -24,9 +24,9 @@ const ImplementedChangesList: React.FC<ImplementedChangesListProps> = ({ changes
         <>
           [<Link to={`${routes.PROJECTS}/${wbsPipe(ic.wbsNum)}`}>{wbsPipe(ic.wbsNum)}</Link>]{' '}
           <OverlayTrigger
-            placement="right"
+            placement={`${window.innerWidth < 600 ? 'top' : 'right'}`}
             overlay={
-              <Tooltip id="tooltip">
+              <Tooltip id="tooltip" style={{ maxWidth: '170px' }}>
                 {fullNamePipe(ic.implementer)} - {datePipe(ic.dateImplemented)}
               </Tooltip>
             }


src/frontend/src/pages/ChangeRequestDetailPage/ImplementedChangesList.tsx
```diff
@@ -6,15 +6,58 @@
...
+  const [bodyWidth, setBodyWidth] = useState<number>(window.document.body.offsetWidth);
+
+  window.document.body.addEventListener('resize', () => {
+    setBodyWidth(window.document.body.offsetWidth);
+  });
+
+  function useWindowSize() {
+    const [innerWidth, setInnerWidth] = useState<number>(0);
+    function determinePosition() {
+      return window.innerWidth <= bodyWidth ? 'top' : 'right';
```
jonathan-chen10, line 29: Looks like you're comparing current window width to window width upon loading. So for me that means tooltips by default appear on top (`(window.innerwidth === bodyWidth) -> 'top'`) unless I shrink the page with Ctrl/Cmd+minus. We want the tooltip to display `right` always unless the tooltop goes beyond the width of the screen.

src/frontend/src/pages/ChangeRequestDetailPage/ImplementedChangesList.tsx
```diff
@@ -6,15 +6,58 @@
...
 interface ImplementedChangesListProps {
   changes: ImplementedChange[];
   overallDateImplemented?: Date;
 }
 
 const ImplementedChangesList: React.FC<ImplementedChangesListProps> = ({ changes, overallDateImplemented }) => {
+  const [bodyWidth, setBodyWidth] = useState<number>(window.document.body.offsetWidth);
+
+  window.document.body.addEventListener('resize', () => {
+    setBodyWidth(window.document.body.offsetWidth);
```
jonathan-chen10, line 23: I tried resetting the width, and it didn't seem to call setBodyWidth in here. So bodyWidth was fixed at loading time. Not sure if that's what you intended.

jonathan-chen10 (changes requested):
Left some comments for ImplementedChangesList. Not sure it's working properly on my end. I didn't check ChangesList yet, but the same comments should apply.

src/frontend/src/pages/ChangeRequestDetailPage/ImplementedChangesList.tsx
zachnorman02: Sorry, just getting around to this. Seems like it's no longer working for me either so maybe it working a few weeks ago was a fluke. Since there is also word-wrapping right now, not sure if it would be worth pushing the word wrapping + always keeping it at the top to prevent overflow, and then trying to debug switching the positioning.

src/frontend/src/components/ChangesList.tsx
```diff
@@ -4,19 +4,62 @@
...
 }
 
 const ChangesList: React.FC<ChangesListProps> = ({ changes }) => {
+  const [bodyWidth, setBodyWidth] = useState<number>(window.document.body.offsetWidth);
+
+  window.document.body.addEventListener('resize', () => {
+    setBodyWidth(window.document.body.offsetWidth);
+  });
+
+  function useWindowSize() {
```
z-roth, line 28: lets try to abstract this hook into a separate file

z-roth, line 28: its a nit, but lets try to use arrow functions instead of the function keyword

src/frontend/src/components/ChangesList.tsx
```diff
@@ -4,19 +4,62 @@
...
+      updateTooltipProps();
+      return () =>
+        window.removeEventListener('resize', () => {
+          updateTooltipProps();
+        });
+    }, []);
+    return [innerWidth, position];
+  }
+
+  let [innerWidth, position] = useWindowSize() as [number, 'top' | 'right'];
```
z-roth, line 51: do we need to cast this?

src/frontend/src/components/ChangesList.tsx
```diff
@@ -4,19 +4,62 @@
...
+
+  // https://mui.com/material-ui/react-tooltip/#VariableWidth.tsx
+  const DynamicTooltip = styled(({ className, ...props }: TooltipProps) => (
```
z-roth, line 54: lets break this out into another file as well

src/frontend/src/pages/ChangeRequestDetailPage/ImplementedChangesList.tsx
```diff
@@ -6,16 +6,59 @@
...
+  });
+
+  function useWindowSize() {
```
z-roth, line 27: lets minimize code duplication by breaking this into a new custom hook in a separate file

z-roth (changes requested):

src/frontend/src/components/ChangesList.tsx
zachnorman02: When I delete it, when I set the Position for the tooltip, I get this type error. There's probably a better way or something I'm missing but casting was the only thing that resolved it for me
image

jonathan-chen10 (changes requested):
Thanks for putting so much work into this ticket. I was not expecting it to be this frustrating!

So far here's how it looks to me: Tooltips start on the right side and as your shrink the window they end up on the left (??) side and when you shrink it more it goes back to the right. Looks like the first transition happens when the window gets to small (as expected), not too sure when the second transition occurs.

(and if you end up getting a solution that doesn't use DynamicTooltip remember to delete the component)

src/frontend/src/components/ChangesList.tsx
```diff
@@ -29,14 +29,27 @@ const ChangesList: React.FC<ChangesListProps> = ({ changes }) => {
           ]{' '}
           <Tooltip
```
z-roth, line 30: could we abstract this into a new custom component?

z-roth (changes requested):

jonathan-chen10 (changes requested):
Very nice and clean solution!

I find that sometimes when I shrink to page to a certain width, I can get the popper to flicker between top and right orientations. Maybe [this](https://tinyurl.com/3bvxd8ed) can help (play around with the offset numbers).

src/frontend/src/components/ChangesList.tsx
```diff
@@ -26,21 +27,11 @@ const ChangesList: React.FC<ChangesListProps> = ({ changes }) => {
...
-            }
-            placement="right"
-            arrow
-          >
-            <Typography component="span">{ic.detail}</Typography>
-          </Tooltip>
+          ]
+          <DynamicTooltip
+            title={`${fullNamePipe(ic.implementer)} - ${datePipe(ic.dateImplemented)}`}
+            children={<Typography component="span">{ic.detail}</Typography>}
```
z-roth, line 32: instead of declaring children as a prop, you declare children by making them child components of the parent. here, you would do:

`<DynamicTooltip ...>children go here</DynamicTooltip>`

src/frontend/src/pages/ChangeRequestDetailPage/ImplementedChangesList.tsx
```diff
@@ -24,25 +25,20 @@ const ImplementedChangesList: React.FC<ImplementedChangesListProps> = ({ changes
       <List>
         {changes.map((ic, idx) => (
           <ListItem key={idx}>
-            <Tooltip
-              id="tooltip"
-              title={
+            <DynamicTooltip
+              title={`${fullNamePipe(ic.implementer)} - ${datePipe(ic.dateImplemented)}`}
+              children={
```
z-roth, line 30: same here

z-roth (changes requested):
great work! finish line is in sight (haha see what i did there)

also, lets clean up the linting errors and we should be good to go!

src/frontend/src/components/ChangesList.tsx
```diff
@@ -26,21 +26,10 @@ const ChangesList: React.FC<ChangesListProps> = ({ changes }) => {
...
-              </>
-            }
-            placement="right"
-            arrow
-          >
-            <Typography component="span">{ic.detail}</Typography>
-          </Tooltip>
+          ]
+          <DynamicTooltip title={`${fullNamePipe(ic.implementer)} - ${datePipe(ic.dateImplemented)}`}>
+            {<Typography component="span">{ic.detail}</Typography>}
```
jonathan-chen10, line 31: Don't think the curly braces around this are needed

src/frontend/src/components/DynamicTooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+            {
+              name: 'flip',
+              options: {
+                fallbackPlacements: ['top', 'bottom']
+              }
+            },
+            {
+              name: 'offset',
+              options: {
+                // I think this should allow a plain value as well, not functions
```
jonathan-chen10, line 26: I think we can get rid of this

src/frontend/src/components/DynamicTooltip.tsx
```diff
@@ -0,0 +1,40 @@
...
+                offset: [0, -10] // distance: 10px
```
jonathan-chen10, line 27: It still can get jittery, idk :( try searching "flicker" or "jitter" or something in the [popper issues](https://github.com/floating-ui/floating-ui/issues) and see if anything works. If nothing works then it's not too big of a problem tbh

jonathan-chen10 (changes requested):

src/frontend/src/components/DynamicTooltip.tsx
```diff
@@ -0,0 +1,40 @@
+import { Tooltip, Typography } from '@mui/material';
```
jonathan-chen10, line 1: Might want to put "This file is part of NER's FinishLine and licensed under GNU AGPLv3." on top like some of the other components (copy the comment from them). Not sure if it's important or not but good to have

src/frontend/src/components/DynamicTooltip.tsx
zachnorman02: Seems like at the exact boundary where it switches from right to top is where it can't decide and flickers. I'm guessing the offset just moved where that happens by 10px. I'll look into it some more and see if I can find a solution.

z-roth:
Have you tried experimenting with the `overflow: hidden` property to fix the flickering? If you haven't, I think that might work. It's ok if it goes a *little* off screen.

z-roth:
@zachnorman02 could you do it on just the container that the popper is contained on? would that work?

anthonybernardi (approved):
very cool! nice work getting through this one

Issue #368, opened 2022-10-20
Teams - Set Team Members Endpoint


joshiarnav:
### Description

Create a post endpoint to handle editing team members in a team (see #362).

### Acceptance Criteria

Create an endpoint that handles editing team members as mentioned above. The route for this would be ```/set-members``` on the specific team page (so ```/teams/:teamId/set-memberse```) as a post request. 

The post body parameters needed are an array of ```userId``` (array of integers) and ```teamId``` (integer).

This endpoint should throw an error if any of the users do not exist or if the the submitter does not have the permissions of admin/app_admin/leader of the team

### Proposed Solution

Add the route to the teams routes file and the function to the team controller file.

Add the tests to a teams test file (please create it if it does not exist).

PR #501, opened 2022-12-20
#368 set team members endpoint

callmejim1226:
## Changes

Added Team's set members endpoint to backend team routes and controllers with its test after merging team controller refactor.  

Also, added set team members endpoint to the team API and hooks so set members endpoint could be used from our frontend. 

## Notes

Currently, found the potential issue on team's prisma type while making the test. Currently, my test in teams.test.ts fails b/c from `expect(res).toStrictEqual(updatedSharedTeam1);` line, the res is missing **members** field since `jest.spyOn(prisma.team, 'update').mockResolvedValue(prismaTeam1);` gives a prisma type team object that has no members field. Therefore, when prismaTeam1 gets called to the teamTransformer, it will be missing the **members** field and fails `expect(res).toStrictEqual(updatedSharedTeam1)` test. The potential solution that I thought was to add **members** field in team's prisma type but let me know if there's another way to solve this issue. 

Also, I removed userId from the post body request that was supposed to represent the submitter's userId since we now have getCurrentUser() utility function made. 

## Test Cases

- Case A
- test if setTeamMembers works by giving appropriate submitter, teamId, and userIds as an array of integer. 
- Result:
- there's currently an issue with this as noted in **Notes** section

## Screenshots

**Get certain teamId's team object (Before):**
#368-team-set-members-ss-before

**Post request to update corresponding team's members with userIds: [4,5] (After):**
#368-team-set-members-ss-after

## To Do

- [x] Fix the test for this endpoint

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes # (issue #)


src/backend/src/utils/users.utils.ts
```diff
@@ -52,3 +53,27 @@ export const rankUserRole = (role: Role) => {
...
   }
 };
+
+/**
+ * Produce a array of User with given userIds
+ * @param userIds array of userIds as an array of integers
+ * @returns array of User with type = User[]
+ * @throws if any user does not exist
+ */
+export const getUsers = async (userIds: number[]) => {
```
anthonybernardi, line 63: this is a lot cleaner and should work:
```
  const users = await prisma.user.findMany({
    where: { userId: { in: userIds } }
  });

  if (users.length !== userIds.length) {
    const prismaUserIds = users.map((user) => user.userId);
    const missingUserIds = userIds.filter((id) => !prismaUserIds.includes(id));
    throw new HttpException(404, `${`User(s) with the following ids not found: ${missingUserIds.join(', ')}`}`);
  }

  return users;
```

src/backend/tests/teams.test.ts
```diff
@@ -43,4 +45,36 @@ describe('Teams', () => {
...
+      expect(prisma.team.update).toHaveBeenCalledWith({
+        where: { teamId },
+        data: {
+          members: {
+            set: userIds
+          }
+        },
+        ...teamQueryArgs
+      });
+      expect(res).toStrictEqual(updatedSharedTeam1);
```
anthonybernardi, line 77: beacuse you mocked prisma.team.update to return `prismaTeam1`, that's what res should be. it also means you don't need updatedSharedTeam1 at all which is good becuase we want to limit our test data

src/frontend/src/apis/teams.api.ts
```diff
@@ -18,3 +18,9 @@ export const getSingleTeam = (id: string) => {
     transformResponse: (data) => JSON.parse(data)
   });
 };
+
+export const editSingleTeam = (id: string, payload: any) => {
```
anthonybernardi, line 22: this should be `setTeamMembers` because we'll be having other endpoints that involve editing a tema like editing the description

src/frontend/src/hooks/teams.hooks.ts
```diff
@@ -16,6 +16,21 @@ export const useAllTeams = () => {
 export const useSingleTeam = (teamId: string) => {
   return useQuery<Team, Error>(['temas', teamId], async () => {
     const { data } = await getSingleTeam(teamId);
-    return data
+    return data;
   });
-}
+};
+export const useEditSingleTeam = (teamId: string) => {
```
anthonybernardi, line 22: linebreak above and also see my other comment about naming

src/frontend/src/hooks/teams.hooks.ts
```diff
@@ -16,6 +16,21 @@ export const useAllTeams = () => {
...
+  const queryClient = useQueryClient();
+  return useMutation<{ message: string }, Error, any>(
+    ['teams', 'edit'],
+    async (teamPayload: any) => {
+      const { data } = await editSingleTeam(teamId, teamPayload);
+      return data;
+    },
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries(['teams']);
```
anthonybernardi, line 32: LMAO the other queries use the field "temas" instead of "teams" so can you fix those so this works 😂 i think i probably wrote that woops

src/frontend/src/hooks/teams.hooks.ts
```diff
@@ -16,6 +16,21 @@ export const useAllTeams = () => {
...
-    return data
+    return data;
   });
-}
+};
+export const useEditSingleTeam = (teamId: string) => {
+  const queryClient = useQueryClient();
+  return useMutation<{ message: string }, Error, any>(
+    ['teams', 'edit'],
+    async (teamPayload: any) => {
```
anthonybernardi, line 26: teamPayload should be a int[] right

src/frontend/src/utils/urls.ts
```diff
@@ -54,6 +54,7 @@ const changeRequestCreateProposeSolution = () => `${changeRequestsCreate()}/prop
 /**************** Teams Endpoints ****************/
 const teams = () => `${API_URL}/teams`;
 const teamsById = (id: string) => `${teams()}/${id}`;
+const teamEdit = (id: string) => `${teams()}/${id}/set-members`;
```
anthonybernardi, line 57: `teamsSetMembers`

anthonybernardi (changes requested):
awesome work! lmk if you have questions

src/backend/tests/teams.test.ts
callmejim1226: I had to put sharedTeam1 to be equal to res since updateSingleTeam returns the team object after transforming it. In other words, it returns team obj with members and projects fields.

src/backend/src/controllers/teams.controllers.ts
```diff
@@ -23,4 +24,19 @@ export default class TeamsController {
...
     }
   }
+
+  static async setMembers(req: Request, res: Response, next: NextFunction) {
+    try {
+      const { userIds } = req.body;
+      const submitter = await getCurrentUser(res);
+
+      // update the team with the input fields
+      const updateTeam = await TeamsService.updateSingleTeam(submitter, req.params.teamId, userIds);
```
anthonybernardi, line 34: the service function name should be the same as the controller name (both should be setMembers)

src/backend/src/services/teams.services.ts
```diff
@@ -32,4 +34,47 @@ export default class TeamsService {
...
+  static async updateSingleTeam(submitter: User, teamId: string, userIds: number[]): Promise<Team> {
+    // find and vertify the given teamId exist
+    const team = await prisma.team.findUnique({
+      where: { teamId },
+      ...teamQueryArgs
+    });
+
+    if (!team) throw new NotFoundException('Team', teamId);
+    if (submitter.role !== Role.ADMIN && submitter.role !== Role.APP_ADMIN && submitter.userId !== team.leaderId)
+      throw new AccessDeniedException('No priviledge to update team members');
```
anthonybernardi, line 55: the AccessDeniedException automatically says `Access Denied` so if we're adding the message then it should explain why

so maybe something like `you must be an admin or the team lead to update the members!`

src/backend/src/utils/users.utils.ts
```diff
@@ -52,3 +53,23 @@ export const rankUserRole = (role: Role) => {
       return 1;
   }
 };
+
+/**
+ * Produce a array of User with given userIds
+ * @param userIds array of userIds as an array of integers
+ * @returns array of User with type = User[]
```
anthonybernardi, line 60: instead of saying "with type = user[]" here, add a type annotation to the function itself of `User[]`

src/backend/src/utils/users.utils.ts
```diff
@@ -52,3 +53,23 @@ export const rankUserRole = (role: Role) => {
...
+ */
+export const getUsers = async (userIds: number[]) => {
+  const users = await prisma.user.findMany({
+    where: { userId: { in: userIds } }
+  });
+
+  if (users.length !== userIds.length) {
+    const prismaUserIds = users.map((user) => user.userId);
+    const missingUserIds = userIds.filter((id) => !prismaUserIds.includes(id));
+    throw new HttpException(404, `${`User(s) with the following ids not found: ${missingUserIds.join(', ')}`}`);
```
anthonybernardi, line 71:
```suggestion
    throw new HttpException(404, `User(s) with the following ids not found: ${missingUserIds.join(', ')}`);
```

src/backend/tests/teams.test.ts
```diff
@@ -43,4 +45,36 @@ describe('Teams', () => {
     expect(prisma.team.findUnique).toHaveBeenCalledTimes(1);
     expect(prisma.team.findUnique).toHaveBeenCalledWith({ where: { teamId }, ...teamQueryArgs });
   });
+
+  describe('editTeam', () => {
```
anthonybernardi, line 49: should change describe name to setMembers or whatever the name becomes

src/backend/tests/teams.test.ts
```diff
@@ -43,4 +45,36 @@ describe('Teams', () => {
...
+    test('setTeamMembers works', async () => {
+      jest.spyOn(prisma.team, 'findUnique').mockResolvedValue(prismaTeam1);
+      jest.spyOn(prisma.team, 'update').mockResolvedValue(prismaTeam1);
+      jest.spyOn(userUtils, 'getUsers').mockResolvedValue([superman, wonderwoman]);
```
anthonybernardi, line 53: can we add a test case for this getUsers function where multiple aren't found and it throws the right stuff

src/backend/src/controllers/teams.controllers.ts
anthonybernardi: actually let's go with `setTeamMembers` for both. that's what the api and hook uses and i think that's the most clear

src/frontend/src/apis/teams.api.ts
```diff
@@ -18,3 +18,9 @@ export const getSingleTeam = (id: string) => {
     transformResponse: (data) => JSON.parse(data)
   });
 };
+
+export const setTeamMembers = (id: string, payload: any) => {
```
anthonybernardi, line 22: let's rename the payload to be `userIds` and make it type `number[]`

then instead of `{ ...payload }` we can just do `{ userIds }

anthonybernardi (changes requested):
awesome work! looks great, mostly just stylistic stuff here and adding a test

src/backend/tests/teams.test.ts
```diff
@@ -46,7 +47,14 @@ describe('Teams', () => {
     expect(prisma.team.findUnique).toHaveBeenCalledWith({ where: { teamId }, ...teamQueryArgs });
   });
 
-  describe('editTeam', () => {
+  describe('setTeamMembers', () => {
+    test('setTeamMembers members not found', async () => {
+      jest.spyOn(prisma.team, 'findUnique').mockResolvedValue(prismaTeam1);
+
+      const res = await TeamsService.setTeamMembers(flash, sharedTeam1.teamId, [122, 55]);
+      expect(res).toThrow(new HttpException(404, 'User(s) with the following ids not found: 122, 55'));
```
callmejim1226, line 55: For the getUsers function, I tried to test the case when the given userIds do not exist, but I wasn't sure how to test on expecting HttpException to be thrown. (This line test fails) I would appreciate any feedback!

anthonybernardi (approved):
awesome!!

Issue #462, opened 2022-11-26
Transition to MUI - Add Consistent Typography Tags


jonathan-chen10:
### Important Note

Make sure to branch from [#142-MUI-migration](https://github.com/Northeastern-Electric-Racing/FinishLine/tree/%23142-MUI-migration) and request merge back to [#142-MUI-migration](https://github.com/Northeastern-Electric-Racing/FinishLine/tree/%23142-MUI-migration) for this ticket. We are doing this work on a feature branch because it affects the whole codebase and is prone to breaking things, so a feature branch allows us to test the code without breaking develop.

This ticket doesn't require changes to `package.json` or `yarn.lock`.

### Description

We want to wrap as much as possible of the text on the website with MUI `<Typography>` tags to allow the text to be up-to-date with the MUI themes. 

### Acceptance Criteria

All text in MUI `Box` and `Grid` are surrounded by `<Typography></Typography>` tags. See [WorkPackageDetails.tsx](https://github.com/Northeastern-Electric-Racing/FinishLine/blob/%23142-MUI-migration/src/frontend/src/pages/WorkPackageDetailPage/WorkPackageViewContainer/WorkPackageDetails.tsx) as an example.

### Proposed Solution

Find as much non-`Typography` text as you can and wrap them in the tags. A good place to start is by searching for all instances of MUI's `<Box>` and `<Grid>` and making sure all text inside of them is `<Typography>` and not raw HTML or raw `{ js }` strings.

You may need to `import Typography from '@mui/material/Typography'` or `import { Typography } from '@mui/material/Typography'` if it has not been done already. Be consistent with the import statement at the top of each file.

PR #513, opened 2022-12-27
#462 All Box/Grid texts wrapped in <Typography>

nezamj03:
Added necessary imports and searched everywhere and wrapped (where I thought would need) texts in Typography tags.


src/frontend/src/components/EditableTextInputList.tsx
```diff
@@ -9,6 +9,7 @@ import Button from '@mui/material/Button';
 import TextField from '@mui/material/TextField';
 import DeleteIcon from '@mui/icons-material/Delete';
 import IconButton from '@mui/material/IconButton';
+import { Typography } from '@mui/material';
```
jamescd18, line 12: why switch import syntax? if not intentional, switch to 

`import Typography from '@mui/material/Typography';`

jamescd18: see this MUI documentation for reasoning: https://mui.com/material-ui/guides/minimizing-bundle-size/

src/frontend/src/components/EditableTextInputList.tsx
```diff
@@ -94,37 +95,39 @@ const EditableTextInputList: React.FC<EditableTextInputListProps> = ({
...
-                  disabled={disabledItems && disabledItems[index]}
-                  onClick={() => removeButtonOnClick(index)}
-                >
-                  <DeleteIcon />
-                </IconButton>
-              )
-            }}
-          />
+          <Typography>
+            <TextField
```
jamescd18, line 99: does TextField really need to be wrapped in Typography? I thought Typography was just for bare text...

src/frontend/src/components/ReactHookEditableList.tsx
```diff
@@ -1,4 +1,4 @@
-import { Grid, Button, IconButton, TextField } from '@mui/material';
+import { Grid, Button, IconButton, TextField, Typography } from '@mui/material';
 import DeleteIcon from '@mui/icons-material/Delete';
```
jamescd18, line 2: IMO we should have consistency across files and within imports in the same file...

src/frontend/src/pages/ChangeRequestDetailPage/ActivationDetails.tsx
```diff
@@ -17,25 +17,33 @@ const ActivationDetails: React.FC<ActivationDetailsProps> = ({ cr }) => {
...
           {datePipe(cr.startDate)}
         </Grid>
         <Grid item xs={6} md={2}>
-          <b>Project Manager </b>
+          <Typography>
+            <b>Project Manager </b>
+          </Typography>
         </Grid>
         <Grid item xs={6} md={2}>
           {fullNamePipe(cr.projectManager)}
```
jamescd18, line 41: pretty sure these pipes output strings and thus also need to be wrapped in Typography

src/frontend/src/pages/ChangeRequestDetailPage/ProposedSolutionForm.tsx
```diff
@@ -190,7 +190,7 @@ const ProposedSolutionForm: React.FC<ProposedSolutionFormProps> = ({
                 form="individual-proposed-solution-form"
                 sx={{ textTransform: 'none', fontSize: 16, marginTop: 3 }}
               >
-                Add
+                <Typography>Add</Typography>
```
jamescd18, line 193: same thing, I don't think this is actually necessary if the button component the text is inside is a MUI component

src/frontend/src/pages/InfoPage.tsx
```diff
@@ -22,13 +22,17 @@ const InfoPage: React.FC = () => {
...
-              description={'Glossary Document'}
-              link={'https://docs.google.com/document/d/1_kr7PQxjYKvBTmZc8cxeSv5xx0lE88v0wVXkVg3Mez8/edit?usp=sharing'}
-            />
+            <Typography>
+              <ExternalLink
+                icon={faScroll}
+                description={'Glossary Document'}
+                link={'https://docs.google.com/document/d/1_kr7PQxjYKvBTmZc8cxeSv5xx0lE88v0wVXkVg3Mez8/edit?usp=sharing'}
+              />
+            </Typography>
```
jamescd18, line 31: we should follow principles of abstraction and put the typography in the external link component where it needs to be, rather than wrapping every usage of the external link component

src/frontend/src/pages/ProjectDetailPage/ProjectEdit/ProjectEditContainer.tsx
```diff
@@ -148,15 +148,17 @@ const ProjectEditContainer: React.FC<ProjectEditContainerProps> = ({ project, ex
...
-          <ReactHookTextField
-            name="summary"
-            control={control}
-            sx={{ width: '50%' }}
-            label="Summary"
-            multiline={true}
-            rows={5}
-            errorMessage={errors.summary}
-          />
+          <Typography>
```
jamescd18, line 151: same idea as external link component. if the ReactHookTextField component definitely needs usage of typography, we would do that inside the component, not wrapping it

src/frontend/src/pages/ProjectDetailPage/ProjectEdit/ProjectEditContainer.tsx
```diff
@@ -198,10 +202,10 @@ const ProjectEditContainer: React.FC<ProjectEditContainerProps> = ({ project, ex
 
       <Box textAlign="center" sx={{ my: 2 }}>
         <Button variant="contained" color="success" type="submit" sx={{ mx: 2 }}>
-          Submit
+          <Typography>Submit</Typography>
```
jamescd18, line 205: same, don't think these are needed

jamescd18 (changes requested):
Love the thoroughness and initiative here! A few things on feedback since I see a lot of the same popping up:

1. I'm fairly certain from my understanding of MUI and the Typography component, we should only be using it to wrap bare text in our app, typically not text within other MUI components. 

Layout components like Box and Grid don't apply MUI styling to bare text within them, thus needing the Typography component. But I'm fairly certain that text in a MUI button is already going to be formatted with the proper button text styling by MUI. Here I see a lot of instances that I don't believe actually need the Typography component. I could be wrong here though, so please do some Googling, reading MUI documentation, and experimenting to prove me wrong if you're sure I'm wrong. Always happy to hear more and learn about how MUI works if I'm mistaken.

2. Abstraction and creation of our own reusable react components provides DRY benefits.

In case you haven't heard it before, DRY stands for Don't Repeat Yourself, and is a core programming principle. Where possible, avoid duplicating the exact same code by using abstraction and stuff. Here, ExternalLink and ReactHookTextField are both components that we wrote ourselves and have used in numerous different spots in the app. If they require Typography (which I don't think ReactHookTextField does, but ExternalLink might), then we would be able to write many many fewer lines by simply adding the Typography into those component's definitions instead of wrapping every usage of those components with a Typography.

3. Import statement consistency between default and named imports.

If you're not familiar with TypeScript import statements, there's named imports and default imports that you'll want to familiarize yourself with being able to recognize. When importing content from MUI, I think we should be consistent with our import syntax wherever possible, whether that's always using named imports, or always using default imports. I defer to Anthony and Reid for info on what has been decided, but I'll note my personal preference is for always using named imports due to the benefits in minimizing bundle size: https://mui.com/material-ui/guides/minimizing-bundle-size/

Let me know if there's any thoughts, questions, or concerns about this PR feedback. Happy to chat on Slack any time. Keep up the great work!

src/frontend/src/components/EditableTextInputList.tsx
nezamj03: gotcha -- 

1. love the point about DRY. ill go refactor where i can
2. i did some experimenting and wrapping other components in typography then setting some attributes of the typography does change the text
3. i have it auto import typography so thats why theres some inconsistency probably. i'll change to be consistent where i can -- i think i am going to make them all like like 12 here though, because most of the files follow this format.

nezamj03: *just read over the documentation for bundle sizing, and it makes sense to actually have imports more specific - so i will move forward like that

nezamj03: Screen Shot 2022-12-27 at 1 57 55 AM

a button's text can also be customized using typography. should i keep a typography tag within that or omit it? we certainly like the way the buttons look without it, but should i include for future changes? or i would assume if we want to make changes to a set of buttons in the future we would make some abstraction for a styled button, so maybe i shouldn't include?

PR #514, opened 2022-12-27
#373 newProject accounts for team

### LINKED ISSUE #373: Teams - Modify Create Project to include Teams
### Description

Creating a project currently does not take into account what team the project should be assigned to (as it is a new feature) so this needs to be remedied.

### Acceptance Criteria

- Change route to perform checks on team/teamId if necessary
- Change newProject function in projects controller file to take in a team and send a team to prisma

### Proposed Solution

The acceptance criteria describe the solution essentially; when a new project is created it should now reflect what team the project is assigned to (there should be a ```team``` and ```teamId``` assigned in the project creation). Ensure that this change propogates all the way so all new projects include a team.

### Mocks

_No response_

nezamj03:
## Changes

Made a check to see if the request body includes teamId. If it does not include teamId, a project is similarly created without any teamId, but if it is provided, it verifies for a valid teamId and if it is not, it will send a 404. if it is, it will create the project with the teamId. 

## Notes

n/a

## Test Cases

passed all tests

## To Do

require a teamId in order to not check for an unprovided one.. transition to requiring teamIds

Closes #373

### DIFF
--- a/src/backend/src/controllers/projects.controllers.ts
+++ b/src/backend/src/controllers/projects.controllers.ts
@@ -54,19 +54,12 @@ export const newProject = async (req: Request, res: Response) => {
 
   await validateChangeRequestAccepted(req.body.crId);
 
-  // check for valid team
-  const team = await prisma.team.findUnique({ where: { teamId: req.body.teamId } });
-  if (!team) {
-    return res.status(404).json({ message: `team with id ${req.body.teamId} not found.` });
-  }
-
   // create the wbs element for the project and document the implemented change request
   const maxProjectNumber = await getHighestProjectNumber(req.body.carNumber);
   const createdProject = await prisma.wBS_Element.create({
     data: {
       carNumber: req.body.carNumber,
       projectNumber: maxProjectNumber + 1,
-      teamId: req.body.teamId,
       workPackageNumber: 0,
       name: req.body.name,
       project: { create: { summary: req.body.summary } },
--- a/src/backend/src/controllers/projects.controllers.ts
+++ b/src/backend/src/controllers/projects.controllers.ts
@@ -54,8 +54,6 @@ export const newProject = async (req: Request, res: Response) => {
 
   await validateChangeRequestAccepted(req.body.crId);
 
-  // check for valid team
-
   if (req.body.teamId !== undefined) {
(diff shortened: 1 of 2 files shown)


src/backend/src/controllers/projects.controllers.ts
```diff
@@ -62,6 +62,15 @@ export const newProject = async (req: Request, res: Response) => {
     return res.status(400).json({ message: 'Cannot implement an unreviewed change request' });
   }
 
+  // check for valid team
+
```
anthonybernardi, line 66: delete this line

anthonybernardi (changes requested):
awesome, works perfectly! only thing is I think we should add a test for if the team isn't found. lmk if you need help with that

src/backend/tests/projects.test.ts
```diff
@@ -90,6 +90,14 @@ describe('Projects', () => {
     expect(res.body).toStrictEqual('1.2.3');
   });
 
+  test('newProject fails when unknown team Id provided', async () => {
+    const proj = { ...newProjectPayload, teamId: 'TEST' };
```
anthonybernardi, line 94: you need to add a line here that mocks prisma.team.findunique to return null

anthonybernardi (changes requested):

src/backend/tests/projects.test.ts
```diff
@@ -68,6 +68,17 @@ describe('Projects', () => {
     expect(res.statusCode).toBe(400);
   });
 
+  test('newProject fails when unknown team Id provided', async () => {
+    mockGetChangeRequestReviewState.mockResolvedValue(true);
+    jest.spyOn(prisma.user, 'findUnique').mockResolvedValue({ ...batman, googleAuthId: 'b' });
```
anthonybernardi, line 73: don't need the googleAuthId anymore i think. if you see this in other places can you take it out cuz i think one person did it now everyone thinks we need it lol

anthonybernardi (changes requested):
super tiny thing

src/backend/tests/projects.test.ts
nezamj03: yep

anthonybernardi (approved):
awesome!

PR #517, opened 2022-12-31
#354 Automatic Implementation of start dates for dependent work…

### LINKED ISSUE #354: Initial Propagation of Timeline Impact Changes to Dependent Work Packages
## Overview

Whenever a timeline impact is implemented, propagate timeline shift to all dependent work packages to avoid users needing to implement numerous manual changes when timelines shift.

## Stakeholders

Product Stakeholder: @jamescd18 
Software Stakeholder: @RChandler234 
Reference Users: @zperry02 

## User Story

As a project lead, I want all subsequent work package timelines to adjust appropriately when I implement a timeline impact to the currently active work package without needing to implement additional changes.

## Success Metrics

Success of this feature will be measured by the time it takes a user to fully implement a change request with timeline impact changes on a project with around 5 subsequent work packages. 

Baseline: 3 minutes

## Rollout Plan

- Deploy to production
- Educate users and push them to add dependencies to work packages
- Measure to determine success level

## Out of Scope

tba

## Background / Context

Imagine a project with multiple work packages wherein each work package must be completed before the next one can start. Each work package would then have the previous work package listed in its dependencies list. The first work package is active and all subsequent work packages were set with a start date equal to their previous work package's end date. 

If you were to delay the active work package by adding duration or pushing its start date later, currently within FinishLine, none of the subsequent work packages would change in any way. This would lead to an inaccurate representation of the project's overall end date since the next work package after the active one could no longer be started at its set start date.

## Acceptance Criteria & Mock-ups

When a standard scope change request for work package A is accepted with a proposed solution with a timeline impact greater than zero, find all work packages in the same project that list work package A in their dependencies list and add the timeline impact # of weeks to their start dates. Also record these as implemented changes.

This adjustment to start dates needs to be performed recursively.

Peyton-McKee:
… packages

## Changes

Recursive implementation of dependent work packages and associated timeline statuses and start dates. 

## Notes

Not sure if theres a cleaner way of approaching this, not sure if making prisma calls in the function is smart. 

Screen Shot 2023-03-05 at 10 49 56 PM

Screen Shot 2023-03-05 at 10 50 05 PM

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #354  (issue #)


jamescd18 (changes requested):
This is definitely tricky territory. I was unsure even as I wrote the ticket. I think this is correct, but I have 1 issue and 1 potential suggestion.

Issue: Needs more comments. I've tagged a couple spots, but IMO the entire CR automation code needs a tad more commenting (beyond just what this PR changes). This kind of complex code IMO needs special attention and should be held to a higher standard of commenting to mitigate risk around it becoming misunderstood due to forgetting. Almost every if should be commented, and almost every major line should be commented too. 

Suggestion: we may want to add an int iterator to the `updateDependencies` function that we initialize with a constant for `MAX_DEPENDENCY_RECURSIVE_UPDATE` or something like that. Decrement every time you recur. Not sure if we even have checks to prevent infinite loops in dependencies (if we don't, we definitely should). Or we could just say fuck it this edge case is really unlikely and don't bother doing anything for this. Not sure if this is even a valid concern....

src/backend/src/prisma-query-args/work-package-depedencies.query-args.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Prisma } from "@prisma/client";
+
+const workPackageDependencyQueryArgs = Prisma.validator<Prisma.Work_PackageArgs>()({
+  include: {
+    dependencies: true,
+    wbsElement: true
+  }
+});
+
+export default workPackageDependencyQueryArgs;
```
jamescd18, line 11: newline at end of file missing

src/backend/src/services/change-request.services.ts
```diff
@@ -126,6 +131,9 @@ export default class ChangeRequestsService {
             detail: buildChangeDetail('Duration', String(foundCR.wbsElement.workPackage.duration), String(updatedDuration))
           }
         ];
+
+        await updateDependencies(wpProj.workPackages, foundCR.wbsElement, foundPs.timelineImpact, crId, reviewer);
```
jamescd18, line 135: needs comment

src/backend/src/utils/change-requests.utils.ts
```diff
@@ -49,3 +52,55 @@ export const sendSlackCRReviewedNotification = async (slackId: string, crId: num
...
+ * This function will recursively update the start date of all work packages that depend on the wbs element to be finished
+ */
+export const updateDependencies = async (
+  workPackages: Prisma.Work_PackageGetPayload<typeof workPackageDependencyQueryArgs>[],
+  wbsElement: WBS_Element,
+  timelineImpact: number,
+  crId: number,
+  reviewer: User
+) => {
+  //cycle through all the work packages
```
jamescd18, line 73:
```suggestion
  // cycle through all the work packages
```
this bugs me. generally you should match whatever already exists in the codebase

src/backend/src/utils/change-requests.utils.ts
```diff
@@ -49,3 +52,55 @@ export const sendSlackCRReviewedNotification = async (slackId: string, crId: num
...
+  workPackages.forEach(async (wp) => {
+    //if the work package depends on the wbs element
+    if (wp.dependencies.map((d: WBS_Element) => d.wbsElementId).includes(wbsElement.wbsElementId)) {
+      //update the start date of the work package
+      const copyStartDate = new Date(wp.startDate);
+      const newStartDate = new Date(wp.startDate.setDate(wp.startDate.getDate() + 7 * timelineImpact));
```
jamescd18, line 79: I swear we might have a helper function for this....

jamescd18: might be called end date calculate tho... might perform same function tho...

jamescd18 (changes requested):

src/backend/src/utils/change-requests.utils.ts
```diff
@@ -53,7 +55,56 @@ export const sendSlackCRReviewedNotification = async (slackId: string, crId: num
...
+            update: {
+              changes: {
+                create: change
+              }
+            }
+          }
+        }
+      });
+      // recursively update the work packages that depends on the updated work package
+      updateDependencies(workPackages, wp.wbsElement, timelineImpact, crId, reviewer);
```
anthonybernardi, line 102: so we honestly shouldn't do recursion because of stack limits. im gonna try to keep this simple but here's how i see it:

```
things to note:
- im skipping a lot of type stuff and so things wont be the exact same
- i dont forget exactly how sets work in typescript but it's something like this
- we are gonna need to test the hell out of this


updateDependencies(initialWorkPackage, crid, reviewer, timelineImpact) {
  const seenWorkPackageIds: Set<number> = [initialWorkPackage.id]; // use a set

  const workPackageUpdateQueue = [initialWorkpackage.dependencies];

  while (workPackageUpdateQueue.length !== 0) {
  	const curr = workPackageUpdateQueue.pop(); // take the first one off

  	// something like this
  	if (seenWorkPackageIds.has(curr.id)) {
  	  continue;
  	}

  	seenWorkPackageIds.add(curr.id);

  	// step 1: update curr's startdate 
  	// step 2: create a change for it

  	// step 3: get all of curr's dependencies
  	// step 4: add all of curr's dependencies to workPackageUpdateQueue
  }
}
```

this is pseudocode for how this function should be. This is basically how non-recursive dfs/bfs works depending on where you pop from and append to. it'll be nice and efficient and stuff

the thing with the set is that if we've updated something already, we skip it. eventually we'll run out of work packages because in the largest case we just see every single work package in the database and stuff. (pop removes things from the list so every time that happens it shrinks until it's empty)

anthonybernardi (changes requested):
in response to james' concern about depth of iteration, the best way to solve that is to keep track of what work packages we've seen so far. also instead of recursion we should do it dynamically with for loops and stuff so we dont hit the recursion limit. i wrote some pseudocode below. this also inherently means we won't care if some dumb user set up a cycle i think

anthonybernardi (changes requested):
(putting this in a separate comment for posterity)

so we honestly shouldn't do recursion because of stack limits. im gonna try to keep this simple but here's how i see it:

things to note:
- im skipping a lot of type stuff and so things wont be the exact same
- i dont forget exactly how sets work in typescript but it's something like this
- we are gonna need to test the hell out of this

```
updateDependencies(initialWorkPackage, crid, reviewer, timelineImpact) {
  const seenWorkPackageIds: Set<number> = [initialWorkPackage.id]; // use a set

  const workPackageUpdateQueue = [initialWorkpackage.dependencies];

  while (workPackageUpdateQueue.length !== 0) {
  	const curr = workPackageUpdateQueue.pop(); // take the first one off

  	// something like this
  	if (seenWorkPackageIds.has(curr.id)) {
  	  continue;
  	}

  	seenWorkPackageIds.add(curr.id);

  	// step 1: update curr's startdate 
  	// step 2: create a change for it

  	// step 3: get all of curr's dependencies
  	// step 4: add all of curr's dependencies to workPackageUpdateQueue
  }
}
```

this is pseudocode for how this function should be. This is basically how non-recursive dfs/bfs works depending on where you pop from and append to. it'll be nice and efficient and stuff

the thing with the set is that if we've updated something already, we skip it. eventually we'll run out of work packages because in the largest case we just see every single work package in the database and stuff. (pop removes things from the list so every time that happens it shrinks until it's empty)

src/shared/src/date-utils.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+const addWeeksToDate = (start: Date, weeks: number) => addDaysToDate(start, weeks * 7);
+
+/**
+ * Add the given number of days to the given date and return the outcome.
+ * @param start the start date
+ * @param days the number of days to add to the start date
+ * @returns the outcome of adding the days to the start date
+ */
+const addDaysToDate = (start: Date, days: number) => {
```
jamescd18, line 20: dis vs endDatePipe, if that still exists in the back-end

jamescd18 (commented):
LGTM but I swear we have a similar function. This new date-utils file is better but just noting it

src/shared/src/date-utils.ts
anthonybernardi: i did this because of the naming convention where to me "end Date" is more specific to our business logic and I wanted something more universally useable

anthonybernardi: also the pipes are all frontend for the most part

src/backend/src/services/change-requests.services.ts
```diff
@@ -130,6 +139,13 @@ export default class ChangeRequestsService {
             detail: buildChangeDetail('Duration', String(foundCR.wbsElement.workPackage.duration), String(updatedDuration))
           }
         ];
+
+        // update all the dependencies (and nested dependencies) of this work package so that their start dates reflect the new duration
+        if (foundPs.timelineImpact > 0) {
```
anthonybernardi, line 144: could be negative too right? or do we not want that to update it

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/WorkPackageSummary.tsx
```diff
@@ -83,7 +83,7 @@ const WorkPackageSummary: React.FC<WorkPackageSummaryProps> = ({ workPackage })
             </Grid>
             <Grid item xs={6}>
               <Box display="flex" flexDirection="row">
-                <DetailDisplay label="Dependencies" content={listPipe(workPackage.dependencies, wbsPipe)} paddingRight={1} />
+                <DetailDisplay label="BlockedBy" content={listPipe(workPackage.blockedBy, wbsPipe)} paddingRight={1} />
```
anthonybernardi, line 86: should this be "Blocked By" isntead of one word

anthonybernardi (changes requested):
this is super awesome and works correctly on my computer (migration and functionality)

two small comments and then i also think we should change every instance of the word dependencies ngl even in comments

src/backend/src/services/change-requests.services.ts
Peyton-McKee: Matt says safer is to only delay not push forward, but Casey might disagree.

anthonybernardi (approved):
epic

PR #651, opened 2023-01-24
#223 - Use URL Query Strings for Login Redirect

### LINKED ISSUE #223: Frontend - Better Login Redirect
### Description

From James:

> when I click on a finishline link in Slack on mobile, it will open it in the in-app browser, which never has me logged in already (nor does it ever have NU SSO logged in), so im guaranteed to get bumped to the login page (then google login, then NU SSO). so I always prefer to open it in standalone mobile browser instead, by clicking the button in the bottom right of the in-app browser. the problem here is that the page may load to the login page before i can click that button, meaning that the page being opened into the standalone browser will be the login page, not the page I wanna go to.
theoretically, I think if we made redirect after login just store the soon-to-be URL in a query parameter it would fix this

### Acceptance Criteria

Before being redirected to the login page, the browser should store the intended page to go to in the url like `https://finishlinebyner.com/login?page=projects&value=1.1.0`

After the login is done, it should redirect the user to `https://finishlinebyner.com/projects/1.1.0` by using the query parameters. (I think right now it just goes to what was most recently in your history)

### Proposed Solution

You'll need to poke around with what causes the redirect to the login page in the first place, and in there store the intended page as query parameters right before it redirects.

To deal with query parameters, you can look at how any of the completed tickets have done it in this epic: https://github.com/Northeastern-Electric-Racing/FinishLine/issues/30

### Mocks

_No response_

jonathan-chen10:
## Changes

Modify redirect to login page to use URL query strings ("query args"). Loading `https://finishlinebyner.com/a/b/c?sleep=false&abc=123` without logging in will redirect to `https://finishlinebyner.com/login?page=a&value1=b&value2=c&sleep=false&abc=123`, which will send back to the original page after logging in.

This replaces the `postLoginRedirect` prop for the `Login` component, which is removed.

Hard-coded to ignore a single `/` at the end of a path, so `https://finishlinebyner.com/a/b/c/?sleep=false&abc=123` will redirect to the same login query strings.

## Notes

Because we're using `value1`, `value2`, etc... we can't use these query args anywhere else because then login page might get confused with multiple of the same query and send you to the wrong page after logging in. These are pretty common query string names, so if we use query args a lot we might want to rename these.

## Test Cases

No tests written. Try it manually with the above example.

## To Do

- Maybe we want to use [`URLSearchParams`](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams) interface to generate the login query strings. May or may not make the code cleaner.

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #223

### DIFF
--- a/src/frontend/src/app/AppPublic.tsx
+++ b/src/frontend/src/app/AppPublic.tsx
@@ -37,6 +37,8 @@ const AppPublic: React.FC = () => {
     // otherwise, the user needs to login manually
     // prepare query args to store path after login
     const pathParts: string[] = history.location.pathname.split('/').slice(1);
+    // if the URL ended in an extra '/' we don't want to clutter the query args
+    if (pathParts[pathParts.length - 1] === '') pathParts.pop();
     const asQueryArg = (pathPart: string, idx: number): string => {
       if (idx === 0) return `page=${pathPart}`;
       else return `value${idx}=${pathPart}`;
--- a/src/frontend/src/app/AppPublic.tsx
+++ b/src/frontend/src/app/AppPublic.tsx
@@ -35,10 +35,18 @@ const AppPublic: React.FC = () => {
     }
 
     // otherwise, the user needs to login manually
+    // prepare query args to store path after login
+    const pathParts: string[] = history.location.pathname.split('/').slice(1);
+    const asQueryArg = (pathPart: string, idx: number): string => {
+      if (idx === 0) return `page=${pathPart}`;
+      else return `value${idx}=${pathPart}`;
+    };
+    const pathArgs: string = `?${pathParts.map(asQueryArg).join('&')}`;
     return (
       <Redirect
         to={{
           pathname: routes.LOGIN,
+          search: pathArgs + (history.location.search ? `&${history.location.search.slice(1)}` : ''),
           state: { from: e.location }
         }}
       />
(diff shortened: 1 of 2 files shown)


src/frontend/src/app/AppPublic.tsx
```diff
@@ -35,10 +35,20 @@ const AppPublic: React.FC = () => {
     }
 
     // otherwise, the user needs to login manually
+    // prepare query args to store path after login
+    const pathParts: string[] = history.location.pathname.split('/').slice(1);
+    // if the URL ended in an extra '/' we don't want to clutter the query args
+    if (pathParts[pathParts.length - 1] === '') pathParts.pop();
```
anthonybernardi, line 41: shouldn't this be === '/' ? or am i misreading

src/frontend/src/app/AppPublic.tsx
```diff
@@ -35,10 +35,20 @@ const AppPublic: React.FC = () => {
...
+    const asQueryArg = (pathPart: string, idx: number): string => {
```
anthonybernardi, line 42: pretty sure you could just use `reduce` for this

anthonybernardi (changes requested):
this impressive but nearly unreadable

src/frontend/src/app/AppPublic.tsx
jonathan-chen10: `'/a/b/c/'.split('/')` gives you `['', 'a', 'b', 'c', '']`

anthonybernardi (approved):
awesome! this code is always gonna be complicated but this makes a lot more sense

PR #730, opened 2023-01-26
#667 Removed any type from AppPublic

### LINKED ISSUE #667: [Refactoring] - Remove `any` type from AppPublic
### Description

Remove the use of the `any` type from AppPublic. This may seem very simple, but the point of this is just to have you read through the codebase and understand the types/interfaces used in it, or make new ones that fit the shape of the data being used.

### Notes

### Acceptance Criteria

`any` is not used as a type in AppPublic.

### Proposed Solution

Use `Ctrl + F` to find all instances of any, and replace it with the correct type

aidanroche3:
## Changes

Changed type any to AppRenderInput

Closes #667

### DIFF
--- a/src/frontend/src/app/AppPublic.tsx
+++ b/src/frontend/src/app/AppPublic.tsx
@@ -17,11 +17,11 @@ const AppPublic: React.FC = () => {
 
   const devUserId = localStorage.getItem('devUserId');
 
-  interface appRenderInput {
+  interface AppRenderInput {
     location: string;
   }
 
-  const render = (e: appRenderInput) => {
+  const render = (e: AppRenderInput) => {
     // if logged in, go to authenticated app
     if (auth.user) {
       if (auth.user.defaultTheme && auth.user.defaultTheme.toLocaleLowerCase() !== theme.activeTheme) {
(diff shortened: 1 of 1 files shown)


src/frontend/src/app/AppPublic.tsx
```diff
@@ -17,7 +17,11 @@ const AppPublic: React.FC = () => {
 
   const devUserId = localStorage.getItem('devUserId');
 
-  const render = (e: any) => {
+  interface appRenderInput {
```
anthonybernardi, line 20: interfaces should use pascal case

anthonybernardi (changes requested):
nice!

src/frontend/src/app/AppPublic.tsx
```diff
@@ -17,7 +17,11 @@ const AppPublic: React.FC = () => {
 
   const devUserId = localStorage.getItem('devUserId');
 
-  const render = (e: any) => {
+  interface AppRenderInput {
```
anthonybernardi, line 20: ah sorry didn't realize this before but this should go before the component declaration

anthonybernardi (changes requested):
sorry one more small thing!

anthonybernardi:
Screenshot 2023-01-27 at 1 18 21 PM

also this solution doesn't actually work -- we have to use the type that the package wants

anthonybernardi (approved):
nice!

PR #733, opened 2023-01-26
#549 Fixed color of change requests button on default and on hover

### LINKED ISSUE #549: [Frontend] - New Change Request Button Color
### Browser(s)

Chrome, Other (specify below)

### Platform(s)

Desktop

### Version(s)

production (main)

### Observed Behavior

The color of this button when the cursor is not hovering over it is `#ff0000`, which doesn't match the header.

### Expected Behavior

The color of this button when the cursor is not hovering over it should be `#ef4345`: **NER RED**. If the cursor is hovering over it, only then should it become `#ff0000`.

### Steps to Reproduce

1. Go to https://finishlinebyner.com/change-requests 
2. Observe the color of the button, right-click it and click 'inspect' and note the hex value of `background-color`. It doesn't match the header.

### Screenshots (as needed)

image

Sybaru:
## Changes

Updated the "new change request" button to match the color of NER red and added hover functionality

## Screenshots
image

image

## Checklist

Closes #549

### DIFF
--- a/src/frontend/src/components/NERButton.tsx
+++ b/src/frontend/src/components/NERButton.tsx
@@ -3,9 +3,9 @@ import { Button, styled } from '@mui/material';
 export const NERButton = styled(Button)({
   textTransform: 'none',
   fontSize: 16,
+  borderColor: '#0062cc',
+  boxShadow: 'none',
   '&:hover': {
-    backgroundColor: '#ff0000',
-    borderColor: '#0062cc',
-    boxShadow: 'none'
+    backgroundColor: '#ff0000'
   }
 });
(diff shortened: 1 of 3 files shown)


z-roth (approved):
looks good!

RChandler234 (changes requested):
Can you copy the styling that you're using here into the NERButton Component(just replace what's currently there), and then call the NER button here (should be able to use the same props, just without the styling). Ari's right that this looks better, so I wanna have it update everywhere we're using this type of button as well

Ari-S-123 (approved):

RChandler234 (approved):
LGTM 🔥

PR #797, opened 2023-02-02
#692 - Remove any from user.controllers

### LINKED ISSUE #692: [Refactoring] - Remove `any` type from users.controllers
### Description

Remove the use of the `any` type from users.controllers. This may seem very simple, but the point of this is just to have you read through the codebase and understand the types/interfaces used in it, or make new ones that fit the shape of the data being used.

### Notes

### Acceptance Criteria

`any` is not used as a type in users.controllers.

### Proposed Solution

`Ctrl + F`

sahil102103:
## Changes

I changed all any's to their associated types to have the code become more specific

Closes #692


Peyton-McKee (changes requested):
You have two files from the learning session for some reason, I think you need to get rid of those, also header should not be force unwrapped, if it says its optional, you should check that it exists first and throw an error if it doesnt

src/backend/src/controllers/users.controllers.ts
```diff
@@ -64,14 +64,14 @@ export default class UsersController {
...
-  static async logUserInDev(req: any, res: any, next: NextFunction) {
+  static async logUserInDev(req: Request, res: Response, next: NextFunction) {
     try {
       if (process.env.NODE_ENV === 'production') throw new AccessDeniedException('Cant dev login on production!');
 
       const { userId } = req.body;
       const header = req.headers['user-agent'];
 
-      const user = await UsersService.logUserInDev(userId, header);
+      const user = await UsersService.logUserInDev(userId, header!);
```
Peyton-McKee, line 75: We should check that header exists instead of force unwrapping it

PR #840, opened 2023-02-07
#702 removed any from work-packages-services

### LINKED ISSUE #702: [Refactoring] - Remove `any` type from work-packages.services
### Description

Remove the use of the `any` type from work-packages.services. This may seem very simple, but the point of this is just to have you read through the codebase and understand the types/interfaces used in it, or make new ones that fit the shape of the data being used.

### Notes

### Acceptance Criteria

`any` is not used as a type in work-packages.services.

### Proposed Solution

`Ctrl + F`

harish-sundar:
Changed all types with "any" to actual data type in the work-packages-services file

Closes #702

### DIFF
--- a/src/backend/src/services/work-packages.services.ts
+++ b/src/backend/src/services/work-packages.services.ts
@@ -136,7 +136,7 @@ export default class WorkPackagesService {
       );
     }
 
-    if (dependencies.find((dep: any) => equalsWbsNumber(dep, projectWbsNum))) {
+    if (dependencies.find((dep: WBS_Element) => equalsWbsNumber(dep, projectWbsNum))) {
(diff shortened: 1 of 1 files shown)


Peyton-McKee (dismissed):
LGTM

Ari-S-123 (dismissed):

anthonybernardi (changes requested):
so this is not your fault because dependencies says it's a WBS_Element type but actually dependencies should be of type WbsNumber from shared

so what you'll need to do is fix the input argument of dependencies to be WbsNumber then everything you changed to be that as well

Peyton-McKee (approved):
LGTM

anthonybernardi (approved):
nice!

Issue #831, opened 2023-02-05
Frontend - Empty Login Should Not Redirect to Undefined


jonathan-chen10:
### Browser(s)

Chrome, Firefox, Safari

### Platform(s)

Desktop

### Version(s)

develop

### Observed Behavior

If someone types `https://finishlinebyner.com/` (`http://localhost:3000/`) while logged out, it will redirect to `https://finishlinebyner.com/login?page=undefined` (`http://localhost:3000/login?page=undefined`). This means that after logging in, the site will send you to not the home page but instead `/undefined` which is an error 404.

### Expected Behavior

We want the redirect to end up at the home page after logging in from `https://finishlinebyner.com/`.

Best implementation is probably to change the initial redirect to go to `https://finishlinebyner.com/login` with no query strings. This involves poking around `redirectPathParts` in the AppPublic component. But if you find another way to fix the error 404 that's good too.

### Steps to Reproduce

1. Run the dev environment with `yarn start`.
2. Make sure you're logged out of the dev site.
3. Go to `http://localhost:3000/`. It should redirect to `https://localhost:3000/login?page=undefined`.
4. Log in. You should get an error 404 from `https://localhost:3000/undefined`

### Screenshots (as needed)

![Screen Shot 2023-02-05 at 11 49 44 AM](https://user-images.githubusercontent.com/59806366/216832721-62b48d91-cb5f-4b99-964e-b81ff2dd3540.png)
![Screen Shot 2023-02-05 at 11 49 51 AM](https://user-images.githubusercontent.com/59806366/216832726-92185358-c031-48fc-a2ca-47572ef827ae.png)
![Screen Shot 2023-02-05 at 11 50 01 AM](https://user-images.githubusercontent.com/59806366/216832734-1eaf86fe-d69a-4478-8d37-55fd8b7b40af.png)

PR #869, opened 2023-02-15
#831 Removed redirectQueryArgs so user is not brought to undefined page

walker-sean:
## Changes

I removed the const redirectQueryArgs and removed it from the redirect for users that need to login manually, stopping them from being brought to a 404 page.

## Notes

I don't see another use for redirectQueryArgs, but if there does happen to be another use, deleting it fully could have consequences.

## Screenshots

Screen Shot 2023-02-15 at 12 21 19 AM
rcontent.com/11459
Screen Shot 2023-02-15 at 12 21 53 AM
6410/218940640-4a810957-fe
Screen Shot 2023-02-15 at 12 22 12 AM
b7-4697-8087-90c324d43fd1.png">
Screen Shot 2023-02-15 at 12 22 15 AM
Screen Shot 2023-02-15 at 12 22 20 AM

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes # (issue #)
#831

### DIFF
--- a/src/frontend/src/app/AppPublic.tsx
+++ b/src/frontend/src/app/AppPublic.tsx
@@ -43,14 +43,11 @@ const AppPublic: React.FC = () => {
         (prevArgs: string, pathPart: string, idx: number): string => `${prevArgs}&value${idx + 1}=${pathPart}`,
         `?page=${redirectPathParts[0]}`
       );
-    const redirectQueryArgs =
-      redirectPathQueryArgs + (history.location.search ? `&${history.location.search.slice(1)}` : '');
 
     return (
       <Redirect
         to={{
           pathname: routes.LOGIN,
-          search: redirectQueryArgs,
           state: { from: e.location }
         }}
       />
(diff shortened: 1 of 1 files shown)


src/frontend/src/app/AppPublic.tsx
```diff
@@ -43,14 +43,11 @@ const AppPublic: React.FC = () => {
...
         `?page=${redirectPathParts[0]}`
       );
-    const redirectQueryArgs =
-      redirectPathQueryArgs + (history.location.search ? `&${history.location.search.slice(1)}` : '');
 
     return (
       <Redirect
         to={{
           pathname: routes.LOGIN,
-          search: redirectQueryArgs,
```
jonathan-chen10, line 53: We want to keep this because it embeds the current page into the login URL so after logging in we can get back to the current page.

jonathan-chen10 (changes requested):
Good work so far Sean. I think we can fix this bug specifically without removing the entire login URL redirect feature. For context, in #223 we decided to embed redirect info in the URL like `/login?page=projects` in order for anyone who accesses this link to be able to get to the right page after logging in. 

redirectQueryArgs grabs information about the current page from `history.location` to make it work, but it's not producing quite the right result, which is why the empty path `/`  ends up as `?page=undefined`. Instead of removing this feature altogether, we want to keep it while fixing the specific bug.

walker-sean:
Now, there should be no query strings sent only when login redirects to the home page.

callmejim1226 (approved):
Looks good to me! The app runs successfully without errors and confirmed that no undefined page loads up in homepage after logging out through my local machine. 

Great job!

PR #871, opened 2023-02-16
#800-removed all instances of React.FC

### LINKED ISSUE #800: [Frontend] - Remove React.FC
### Spike Type

Technical

### Goal

Stop using `React.FC` for components. Instead, use

```jsx
const App = ({ message }: AppProps) => <div>{message}</div>;
const App = ({ message }: AppProps): JSX.Element => <div>{message}</div>;
```

### Reason for Spike

Amongst other reasons, `React.FC` provides an implicit definition of children, so components will accept even if they are not supposed to.

For example, in `AppMain.tsx`, there are no type errors with this:

```tsx
const AppMain: React.FC = () => {
  return (
    <AppContext>
      <ToastProvider>
        <BrowserRouter>
         // here
          <AppPublic>
            <h1>Children!</h1>
          </AppPublic>
        </BrowserRouter>
      </ToastProvider>
    </AppContext>
  );
};
```

even though `AppPublic` does not accept children (`const AppPublic: React.FC = () => { ...`)

It's been removed from create react app and replaced with `function App() { ...`

- https://github.com/facebook/create-react-app/pull/8177
- https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/

### Additional notes

_No response_

glacierA:
## Changes

As title says, removed all instances of React.FC in favor of either explicitly listing children as React.ReactNode's or replacing inline-definitions of type React.FC with arrow functions returning JSX.Elements.

Closes #800

### DIFF
--- a/src/frontend/src/app/AppAuthenticated.tsx
+++ b/src/frontend/src/app/AppAuthenticated.tsx
@@ -26,7 +26,7 @@ const styles = {
   }
 };
 
-const AppAuthenticated: React.FC = () => {
+const AppAuthenticated = (): JSX.Element => {
   return (
     <>
       <NavTopBar />
--- a/src/frontend/src/app/AppContext.tsx
+++ b/src/frontend/src/app/AppContext.tsx
@@ -7,7 +7,11 @@ import AppContextAuth from './AppContextAuth';
 import AppContextQuery from './AppContextQuery';
 import AppContextTheme from './AppContextTheme';
 
-const AppContext: React.FC = (props) => {
+type Props = {
+  children?: React.ReactNode;
+};
+
+const AppContext = (props: Props) => {
   return (
     <AppContextQuery>
       <AppContextAuth>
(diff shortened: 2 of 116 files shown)


src/frontend/src/app/AppAuthenticated.tsx
```diff
@@ -26,7 +26,7 @@ const styles = {
   }
 };
 
-const AppAuthenticated: React.FC = () => {
+const AppAuthenticated = (): JSX.Element => {
```
RChandler234, line 29: either add JSX.Element everywhere or don't use it anywhere. We want it to be standard

RChandler234 (changes requested):
ganttpage.tsx, navtopbar.tsx, wbsstatus.tsx, loadingindicator.tsx, DetailDisplay.tsx - make sure all of these pages are also changed over, and make sure that any files in the GanttPackage Folder aren't changed at all, that's just a package Ari found to use for the gantt chart, so I don't wanna change it at all

Ari-S-123 (changes requested):
Lots of merge conflicts.

RChandler234:
Will be worked on again at a later time

Issue #872, opened 2023-02-16
[Backend] - API Schema Types


acd124:
### Spike Type

Technical

### Goal

Find out how to ensure the req.body had the proper type that is guaranteed currently by the express validator when the route is declared. For example, in the controller files the code looks like `const { title, notes... } = req.body` but none of the variables title, notes, or any other parts of the body have a type that isn't any.

### Reason for Spike

The backend types are in some cases inconsistent with reality, and the frontend hooks that use the endpoints have no clue what types are expected. This makes it difficult to figure out what should actually be happening, and introducing any types defeats the point of typescript.

### Additional notes

Try looking for a solution with yup since we already use it in the frontend.

PR #937, opened 2023-02-28
#706: Remove Any Change Request Hook

### LINKED ISSUE #706: [Refactoring] - Remove `any` type from change-requests.api
### Description

Remove the use of the `any` type from change-requests.api. This may seem very simple, but the point of this is just to have you read through the codebase and understand the types/interfaces used in it, or make new ones that fit the shape of the data being used.

### Notes

### Acceptance Criteria

`any` is not used as a type in change-requests.api.

### Proposed Solution

`Ctrl + F`

anthonybernardi:
## Changes

removed the any

## Test Cases

- manually tested that creation still works

closes #706 #686

### DIFF
--- a/src/backend/src/routes/change-requests.routes.ts
+++ b/src/backend/src/routes/change-requests.routes.ts
@@ -51,7 +51,6 @@ changeRequestsRouter.post(
 
 changeRequestsRouter.post(
   '/new/standard',
-  intMinZero(body('submitterId')),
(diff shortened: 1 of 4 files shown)


acd124:
Perhaps move the interface to shared since it's the type that the backend expects. There may be a future use for the type in the backend (see #872)

Peyton-McKee (approved):
WWWWWWWW

PR #970, opened 2023-03-03
#968: Gantt Improvements

### LINKED ISSUE #968: Gantt - jump around improvement
when you scroll over but hten expand one of the projects, the chart jumps back to the present day. we don't want that to happen.

i think the way to do this is to not pass in the viewDate prop and to go into the gantt package and make the default value for the viewDate use state new Date() or something. idk if that makes sense, this will take some research and code digging.

anthonybernardi:
## Changes

kinda went out of scope of my ticket but saw some areas to imrpove the gantt stuff

- no longer snaps around when you expand / close stuff
- the blue line starts like 3 weeks in now which is a little nicer
- deleted a lot of nonsense (or stuff i know we'll never use), yikes to whoever wrote this package originally 
- made the layout match the current gantt chart more

https://docs.google.com/spreadsheets/d/1KMaX7BX4c7TXwEU0zSwl87XUk_U6zp7as5WFV6ToyFM/edit#gid=0

## Screenshots

Screenshot 2023-03-03 at 4 22 52 PM

closes #968


Ari-S-123 (changes requested):
Add dark mode screenshot and one of minimum browser size.

src/frontend/src/pages/GanttPage/GanttPageWrapper.tsx
```diff
@@ -9,25 +9,70 @@ import ErrorPage from '../ErrorPage';
...
-}
+const NO_TEAM = 'No Team';
 
-interface GanttDisplayWorkPackage extends WorkPackage {
-  displayOrder: number;
+interface GanttTask extends Task {
+  teamName: string;
 }
 
+const transformWorkPackageToGanttTask = (workPackage: WorkPackage, teamName: string): GanttTask => {
```
RChandler234, line 31: maybe put these two transformation methods in a utils file, this file is getting pretty long

RChandler234 (changes requested):
Left a couple comments for cleaning up the code. One weird thing rn is that if a team doesn't have any projects that fit the other filter parameters, that team block doesn't show up at all even if it's selected in the team dropdown or it's set to all teams. I think having the name show up with some sort of no applicable projects display under it would be better

anthonybernardi:
updated screenshots:
biggish screen:
Screenshot 2023-03-05 at 10 56 23 AM

smaller screen:
Screenshot 2023-03-05 at 10 58 36 AM


gantt chart basically doesnt work on min size screen so i don thtink it's worth posting

RChandler234 (changes requested):
Was talking to Peyton and he said that the engineers think the progress bars on each of the projects aren't useful. I kinda agree because it implies you got that a certain amount of work done by a certain time which isn't really true. I'd say you should just remove the progress shading for now, and we can look into if we could add some more useful coloring of the projects/ work packages (e.g. coloring work packages by stage) later on

anthonybernardi:
Screenshot 2023-03-05 at 9 34 58 PM

Ari-S-123 (approved):

src/frontend/src/utils/gantt.utils.ts
```diff
@@ -92,3 +140,42 @@ export const sortWbs = (a: { wbsNum: WbsNumber }, b: { wbsNum: WbsNumber }) => {
...
+
+/**
+ * Sort function for the teams on the Gantt chart.
+ * Hardcoding the team names so that we can get the teams in the exact order we want.
+ * However, everything else is not hard coded and allows for more or less teams.
+ * @param a name of team
+ * @param b name of other team
+ * @returns a number for sorting
+ */
+export const sortTeamNames = (a: string, b: string): number => {
```
RChandler234, line 152: This is kinda weird, could we make this a map of values to values, or assign the specific values to constants or smthn. All these numbers in here just feel a lil like magic numbers.

RChandler234 (changes requested):

src/frontend/src/utils/gantt.utils.ts
anthonybernardi: they are definitely magic numbers, it's just forcing the sort function to work a certain way

wym values to values

RChandler234: Yeah, I meant teams to values, my b

PR #1072, opened 2023-03-30
#93 move status logic to back end

### LINKED ISSUE #93: Change Requests - Move Status Logic to Back-End
## Description

Move the business logic of change request status to the back-end. The business logic of change request status should only be in the back-end and the front-end should read from the back-end.

## Background

All change requests have a status. This logic relating to determining the status of a change request currently overlaps with the front-end. As of now, we wrote the logic of change request status in the front-end (mainly cause of laziness), but we should move it to the back-end.

## Business Logic
The status of a change request depends on fields within the change request. The change request will be referred to as `cr`.
- If the cr’s `changes` array is not empty, then the status should always be `IMPLEMENTED`
- If the cr’s `changes` array is empty, the cr’s `accepted` field is true, and the cr’s `dateReviewed` field is not empty, then the status should be `ACCEPTED`
- If the cr’s `changes` array is empty, the cr’s `accepted` field is false, and the cr’s `dateReviewed` field is not empty, then the status should be `DENIED`
- If it doesn't fit any of the above criteria, then the status should be `OPEN`

## Proposed Solution

Move business logic of change request status to the back-end. In other words, the back-end would be responsible for determining the status of a change request. The front-end should read this status from the back-end, and then display it.

- [x] Create the `changeRequestStatus` enum in the `shared` package
  - The enum should have 4 values: `IMPLEMENTED`, `ACCEPTED`, `DENIED`, and `OPEN`
- [x] Add the field to the `ChangeRequest` interface and call it "status"
- [x] Update the `change-requests.stub.ts` test data
- [x] Update the change requests transformer to calculate the status of a cr
  - [x] Write a helper function to do the calculation in shared
  - [x] Utilize the helper function in the `changeRequestTransformer` function
- [x] Update the front-end to properly utilize the status field sent by the back-end
  - [x] Remove code from the front-end that computes which status to display (on table and on individual page)
  - [x] Have the front-end display the value provided by the back-end (on the table and an individual cr page)
  - [x] Update tests in the front-end to ensure they are still passing properly
- [x] manually ensure this new feature works properly and post a screenshot in your pr

zachmarino234:
## Changes

1. Created the status enumeration to keep track of the status of a change request 
2. Created a helper function to calculate the status of the change request
3. Created that helper in a new file in the shared directory
4. Removed the frontend helper that calculated the status
5. Made a new column in the change request table that reports the status of the change request
6. Checked to make sure all tests are still working
7. Removed Herobrine

## Screenshots

image
image
image
image

Closes #93

### DIFF
--- a/src/backend/src/transformers/change-requests.transformer.ts
+++ b/src/backend/src/transformers/change-requests.transformer.ts
@@ -9,7 +9,9 @@ import userTransformer from './user.transformer';
 const changeRequestTransformer = (
   changeRequest: Prisma.Change_RequestGetPayload<typeof changeRequestRelationArgs>
 ): ChangeRequest | StandardChangeRequest | ActivationChangeRequest | StageGateChangeRequest => {
-  const cr = {
+  const status = calculateChangeRequestStatus(changeRequest);
+
+  return {
     // all cr fields
     crId: changeRequest.crId,
     wbsNum: wbsNumOf(changeRequest.wbsElement),
@@ -34,6 +36,7 @@ const changeRequestTransformer = (
       detail: change.detail,
       dateImplemented: change.dateImplemented
     })),
+    status,
     // scope cr fields
     what: changeRequest.scopeChangeRequest?.what ?? undefined,
     why: changeRequest.scopeChangeRequest?.why.map((why) => ({
@@ -59,9 +62,6 @@ const changeRequestTransformer = (
     leftoverBudget: changeRequest.stageGateChangeRequest?.leftoverBudget ?? undefined,
     confirmDone: changeRequest.stageGateChangeRequest?.confirmDone ?? undefined
   };
-
-  const crStatus = calculateChangeRequestStatus(changeRequest);
-  return { ...cr, status: crStatus };
 };
 
 export default changeRequestTransformer;


src/shared/src/backend-supports/change-request-supports.ts
```diff
@@ -0,0 +1,32 @@
...
+import { WbsNumber } from '../types/project-types';
+import { User } from '../types/user-types';
+
+/**
+ * This function calculates the status of a change request.
+ * @param cr An incoming change request
+ * @returns The status of the change request. Can either be Open, Accepted, Denied, or Implemented
+ */
+const calculateStatus = (
+  implementedChanges: {
```
nezamj03, line 11: i think you can change to type ImplementedChange[]

src/shared/src/backend-supports/change-request-supports.ts
```diff
@@ -0,0 +1,32 @@
+import { ChangeRequestStatus } from '../types/change-request-types';
```
nezamj03, line 1: need to add the NER message (look at any other file)

src/backend/src/transformers/change-requests.transformer.ts
```diff
@@ -59,6 +65,9 @@ const changeRequestTransformer = (
     leftoverBudget: changeRequest.stageGateChangeRequest?.leftoverBudget ?? undefined,
     confirmDone: changeRequest.stageGateChangeRequest?.confirmDone ?? undefined
   };
+
+  const crStatus = calculateStatus(cr.implementedChanges, cr.accepted, cr.dateReviewed);
+  return { ...cr, status: crStatus };
```
nezamj03, line 70: this is the only thing that anthony might complain about but i think its fine. the alternate way to do it is to make a const at the top...

nezamj03 (changes requested):

nezamj03 (approved):

src/shared/src/backend-supports/change-request-supports.ts
nezamj03: still need to do this

nezamj03 (changes requested):
need to change to type ImplementedChange[] where noted still

nezamj03 (commented):
lgtm!

nezamj03 (dismissed):
lgtm!

src/shared/src/backend-supports/change-request-supports.ts
```diff
@@ -0,0 +1,28 @@
+/*
```
anthonybernardi, line 1: since this function should only be called by the backend, it should go in the backend change request utils

src/shared/src/backend-supports/change-request-supports.ts
```diff
@@ -0,0 +1,28 @@
...
+ */
+
+import { ChangeRequestStatus, ImplementedChange } from '../types/change-request-types';
+
+/**
+ * This function calculates the status of a change request.
+ * @param cr An incoming change request
+ * @returns The status of the change request. Can either be Open, Accepted, Denied, or Implemented
+ */
+const calculateStatus = (
```
anthonybernardi, line 13: `calculateChangeRequestStatus`

src/shared/src/backend-supports/change-request-supports.ts
```diff
@@ -0,0 +1,28 @@
...
+  implementedChanges: ImplementedChange[],
```
anthonybernardi, line 14: this function expects a lot of frontend CR fields which doesn't make sense since the point is that we're taking a backend CR and making it a frontend CR

so have this function take in the backend change request type (the same type as the input for the transformer) and then this function can just take the whole change request in

src/shared/src/types/change-request-types.ts
```diff
@@ -81,6 +82,13 @@ export enum ChangeRequestReason {
   Other = 'OTHER'
 }
 
+export enum ChangeRequestStatus {
+  Implemented = 'IMPLEMENTED',
```
anthonybernardi, line 86: can we make all of these like `Implemented` `Accepted`

(i know all the others are all caps but it's actually so annoying that they're all caps becuase we have to make them not all caps all the time when we display them lol so this gets us ahead of that)

src/frontend/src/pages/ChangeRequestsPage/ChangeRequestsTable.tsx
```diff
@@ -131,6 +131,12 @@ const ChangeRequestsTable: React.FC = () => {
       filterable: false,
       valueFormatter: (params) => params.value.length,
       maxWidth: 200
+    },
```
anthonybernardi, line 134: im tempted to delete the date implemented column ngl now that we have this

src/backend/src/transformers/change-requests.transformer.ts
```diff
@@ -9,7 +15,7 @@ import userTransformer from './user.transformer';
 const changeRequestTransformer = (
   changeRequest: Prisma.Change_RequestGetPayload<typeof changeRequestRelationArgs>
 ): ChangeRequest | StandardChangeRequest | ActivationChangeRequest | StageGateChangeRequest => {
-  return {
+  const cr = {
```
anthonybernardi, line 18: with my other comment, we wont need this weirdness and you'll just be able to add
`status: calculateChangeRequestStatus(changeRequest)`

anthonybernardi (changes requested):
very nice!

anthonybernardi (approved):
WWWWWWWW

PR #1101, opened 2023-04-13
#676 refactoring remove any type from login logindev and loginpage

### LINKED ISSUE #676: [Refactoring] - Remove `any` type from Login
### Description

Remove the use of the `any` type from Login.

### Acceptance Criteria

`any` is not used as a type in Login.

### Proposed Solution

`Ctrl + F`

### Mocks

_No response_

zmar234:
## Changes

Removed `any` from `Login.tsx` `LoginDev.tsx` and `LoginPage.tsx`. I left `any` for `onFailure` because it needs to take in different errors and in the `GoogleLoginProps` interface in `index.d.ts`, `onFailure` takes in any.

## Notes

Removing `any` from `verifyLogin` required adding a tiny bit of extra code to the function. The only way to replace any was to also include `GoogleResponseOffline` but `getAuthResponse` doesn't take in Offline therefore we needed to confirm that Offline would not be fed to `getAuthResponse`. I worked on this solution with Nezam.

Closes #676

### DIFF
--- a/src/frontend/src/pages/LoginPage/Login.tsx
+++ b/src/frontend/src/pages/LoginPage/Login.tsx
@@ -3,7 +3,7 @@
  * See the LICENSE file in the repository root folder for details.
  */
 
-import { useState } from 'react';
+import { FormEvent, useState } from 'react';
 import { useHistory } from 'react-router';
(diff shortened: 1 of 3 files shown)


nezamj03 (approved):
lgtm

PR #1140, opened 2023-05-04
#671 refactoring remove any type from reacthooktextfield

### LINKED ISSUE #671: [Refactoring] - Remove `any` type from ReactHookTextField
### Description

Remove the use of the `any` type from ReactHookTextField. This may seem very simple, but the point of this is just to have you read through the codebase and understand the types/interfaces used in it, or make new ones that fit the shape of the data being used.

### Notes

### Acceptance Criteria

`any` is not used as a type in ReactHookTextField.

### Proposed Solution

Use `Ctrl + F` to find all instances of any, and replace it with the correct type

zmar234:
## Changes

Removed `any` from `ReactHookTextField.tsx`

## Notes

I kept `any` for the generic types of `Control` and `Omit` because those types are basically rulesets that take in many different types to determine how the react text fields operate.

closes #671

### DIFF
--- a/src/frontend/src/components/ReactHookTextField.tsx
+++ b/src/frontend/src/components/ReactHookTextField.tsx
@@ -18,7 +18,7 @@ interface ReactHookTextFieldProps {
   icon?: SvgIconProps;
   multiline?: boolean;
   rows?: number;
-  endAdornment?: React.ReactNode;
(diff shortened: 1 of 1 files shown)


RChandler234 (approved):
LGTM

PR #1283, opened 2023-08-08
#1027 Create change request input component

### LINKED ISSUE #1027: Change Request Dropdown on WP/Project Page
### Description

Currently, users have to manually enter crs into wps or projects when they want to edit a project or wp from the detail page. This ticket will make it easier for them to select the change request that they want to use.

Make the change request field a dropdown that displays all change requests that were approved in the past 5 days. Each dropdown section should have the crs number, wbsnum, and user submitted.

E.G.
1 - 1.2.0 - Thomas Emrax

### Acceptance Criteria

The CR field on the WP and Project page is now a multiselect autocomplete with the specified crs as options.

### Proposed Solution

Use the autocomplete, you'll probably have to fetch all the crs and just filter them for the dropdown.

### Mocks

_No response_

ninest:
## Changes

Convert change request inputs to dropdowns/autocompletes so users don't have to remember CR IDs and manually type them out

## Screenshots

image

image

image

New UI with label:

image

## To Do

- [ ] autcomplete or dropdown?
- [ ] need to create an API for getting change requests? Or filter them client side?

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #1027

### DIFF
--- a/src/frontend/src/components/ChangeRequestDropdown.tsx
+++ b/src/frontend/src/components/ChangeRequestDropdown.tsx
@@ -1,7 +1,7 @@
 import { MenuItem, Select, SelectChangeEvent } from '@mui/material';
 import { isWithinInterval, subDays } from 'date-fns';
 import { Control, Controller } from 'react-hook-form';
-import { wbsPipe } from 'shared';
+import { AuthenticatedUser, ChangeRequest, wbsPipe } from 'shared';
 import { useAllChangeRequests } from '../hooks/change-requests.hooks';
 import LoadingIndicator from './LoadingIndicator';
 import { useCurrentUser } from '../hooks/users.hooks';
@@ -16,24 +16,7 @@ const ChangeRequestDropdown = ({ control, name }: ChangeRequestDropdownProps) =>
   const { isLoading, data: changeRequests } = useAllChangeRequests();
   if (isLoading || !changeRequests) return <LoadingIndicator />;
 
-  const today = new Date();
-  const fiveDaysAgo = subDays(today, 5);
-
-  const filteredRequests = changeRequests?.filter(
-    (cr) => cr.dateReviewed && cr.accepted && isWithinInterval(cr.dateReviewed, { start: fiveDaysAgo, end: today })
-  );
-
-  // The current user's CRs should be at the top
-  filteredRequests.sort((a, b) => {
-    const isSubmitterAUser = a.submitter.firstName === user.firstName && a.submitter.lastName === user.lastName;
-    const isSubmitterBUser = b.submitter.firstName === user.firstName && b.submitter.lastName === user.lastName;
-
-    if (isSubmitterAUser && isSubmitterBUser) return 0;
-    if (isSubmitterAUser) return -1;
-    if (isSubmitterBUser) return 1;
-
-    return a.crId - b.crId;
-  });
+  const filteredRequests = getChangeRequests(changeRequests, user);
 
   const approvedChangeRequestOptions =
     filteredRequests.map((cr) => ({
@@ -71,5 +54,29 @@ const ChangeRequestDropdown = ({ control, name }: ChangeRequestDropdownProps) =>
     />
   );
 };
+
+// Filter and sort change requests to display in the dropdown
+function getChangeRequests(changeRequests: ChangeRequest[], user: AuthenticatedUser): ChangeRequest[] {
+  const today = new Date();
+  const fiveDaysAgo = subDays(today, 5);
+
+  const filteredRequests = changeRequests.filter(
+    (cr) => cr.dateReviewed && cr.accepted && isWithinInterval(cr.dateReviewed, { start: fiveDaysAgo, end: today })
+  );
+
+  // The current user's CRs should be at the top
+  filteredRequests.sort((a, b) => {
+    const isSubmitterAUser = a.submitter.firstName === user.firstName && a.submitter.lastName === user.lastName;
+    const isSubmitterBUser = b.submitter.firstName === user.firstName && b.submitter.lastName === user.lastName;
+
+    if (isSubmitterAUser && isSubmitterBUser) return 0;
+    if (isSubmitterAUser) return -1;
+    if (isSubmitterBUser) return 1;
+
+    return a.crId - b.crId;
+  });
+
+  return filteredRequests;
+}
 
 export default ChangeRequestDropdown;
--- a/src/frontend/src/components/ChangeRequestDropdown.tsx
+++ b/src/frontend/src/components/ChangeRequestDropdown.tsx
@@ -1,4 +1,4 @@
-import { MenuItem, Select, SelectChangeEvent, dividerClasses } from '@mui/material';
+import { MenuItem, Select, SelectChangeEvent } from '@mui/material';
 import { isWithinInterval, subDays } from 'date-fns';
 import { Control, Controller } from 'react-hook-form';
 import { wbsPipe } from 'shared';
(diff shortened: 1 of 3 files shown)


ninest:
image

src/frontend/src/components/ChangeRequestDropdown.tsx
```diff
@@ -0,0 +1,74 @@
...
+  filteredRequests.sort((a, b) => {
+    const isSubmitterAUser = a.submitter.firstName === user.firstName && a.submitter.lastName === user.lastName;
+    const isSubmitterBUser = b.submitter.firstName === user.firstName && b.submitter.lastName === user.lastName;
+
+    if (isSubmitterAUser && isSubmitterBUser) return 0;
+    if (isSubmitterAUser) return -1;
+    if (isSubmitterBUser) return 1;
+
+    return a.crId - b.crId;
+  });
```
ninest, line 35: Is it worth making a helper function for this?

src/frontend/src/components/ChangeRequestDropdown.tsx
```diff
@@ -0,0 +1,74 @@
...
+          value={value}
+          onChange={(event: SelectChangeEvent<number>) => {
+            const {
+              target: { value }
+            } = event;
+            if (value) onChange(value);
+          }}
+          size={'small'}
+          placeholder={'Change Request Id'}
+          sx={{ width: 200, textAlign: 'left' }}
```
ninest, line 61: A few other Selects had width set like this. Also, I had to add textAlign left because it was aligned to the right by default, not sure why

src/frontend/src/components/ChangeRequestDropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+const ChangeRequestDropdown = ({ control, name }: ChangeRequestDropdownProps) => {
+  const user = useCurrentUser();
+  const { isLoading, data: changeRequests } = useAllChangeRequests();
+  if (isLoading || !changeRequests) return <LoadingIndicator />;
+
+  const today = new Date();
+  const fiveDaysAgo = subDays(today, 5);
+
+  const filteredRequests = changeRequests?.filter(
+    (cr) => !!cr.dateReviewed && isWithinInterval(cr.dateReviewed, { start: fiveDaysAgo, end: today })
```
callmejim1226, line 23: Great job with filtering here.

I would suggest removing !! in front of cr.dateReviewed and leaving it as just cr.dateReviewed since it functions the same and is easier to understand.

Also, from the issue description, it said to display all change requests that were **approved** in the past 5 days, so I would suggest we can add a condition of checking cr.accepted here.

src/frontend/src/components/ChangeRequestDropdown.tsx
```diff
@@ -0,0 +1,75 @@
...
+    const isSubmitterBUser = b.submitter.firstName === user.firstName && b.submitter.lastName === user.lastName;
+
+    if (isSubmitterAUser && isSubmitterBUser) return 0;
+    if (isSubmitterAUser) return -1;
+    if (isSubmitterBUser) return 1;
+
+    return a.crId - b.crId;
+  });
+
+  const options =
```
callmejim1226, line 38: We can rename this variable to approvedCROptions or smth like it to be more specific.

src/frontend/src/components/ChangeRequestDropdown.tsx
callmejim1226: I would say yes. We can create a helper function that accepts whole change requests from line 16 and returns the filtered and sorted crs.

callmejim1226 (changes requested):
Great job on making cr dropdown in proj/wp details page!

This is close to being done :)

I added a few comments on the filtering part, variable definition, and your comment on making a helper function.

src/frontend/src/components/ChangeRequestDropdown.tsx
ninest: I'll make a function in the same file

callmejim1226: As long the name of a function specifically explains what it does it will be fine. 

For this case, we can name it as approvedChangeRequests for example.

src/frontend/src/components/ChangeRequestDropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+  const today = new Date();
+  const fiveDaysAgo = subDays(today, 5);
+
+  const filteredRequests = changeRequests.filter(
+    (cr) => cr.dateReviewed && cr.accepted && isWithinInterval(cr.dateReviewed, { start: fiveDaysAgo, end: today })
+  );
+
+  // The current user's CRs should be at the top
+  filteredRequests.sort((a, b) => {
+    const isSubmitterAUser = a.submitter.firstName === user.firstName && a.submitter.lastName === user.lastName;
```
RChandler234, line 69: submitter.userId === user.userId for both of these instead

src/frontend/src/components/ChangeRequestDropdown.tsx
```diff
@@ -0,0 +1,82 @@
...
+    const isSubmitterBUser = b.submitter.firstName === user.firstName && b.submitter.lastName === user.lastName;
+
+    if (isSubmitterAUser && isSubmitterBUser) return 0;
+    if (isSubmitterAUser) return -1;
+    if (isSubmitterBUser) return 1;
+
+    return a.crId - b.crId;
+  });
```
RChandler234, line 77: Do you really need all this code to sort these properly? I think just returning -1 if a is the same as the current user should work the same

RChandler234 (changes requested):

src/frontend/src/components/ChangeRequestDropdown.tsx
ninest: Do you mean something like

```
if (isSubmitterAUser || isSubmitterBUser) return -1;
else return a.crId - b.crId;
```

ninest:
image

src/frontend/src/components/ChangeRequestDropdown.tsx
```diff
@@ -0,0 +1,81 @@
...
+            >
+              {approvedChangeRequestOptions.map((option) => (
+                <MenuItem key={option.value} value={option.value}>
+                  {option.label}
+                </MenuItem>
+              ))}
+            </Select>
+          </Box>
+        )}
+      />
```
RChandler234, line 76: I'd do it like this
```suggestion
      <FormControl>
        <FormLabel sx={{ alignSelf: 'start' }}>Change Request ID</FormLabel>
        <Controller
          control={control}
          name={name}
          render={({ field: { onChange, value } }) => (
            <Select
              id="cr-autocomplete"
              displayEmpty
              renderValue={(value) => value}
              value={value}
              onChange={(event: SelectChangeEvent<number>) => onChange(event.target.value)}
              size={'small'}
              placeholder={'Change Request Id'}
              sx={{ width: 200, textAlign: 'left' }}
            >
              {approvedChangeRequestOptions.map((option) => (
                <MenuItem key={option.value} value={option.value}>
                  {option.label}
                </MenuItem>
              ))}
            </Select>
          )}
        />
      </FormControl>
```

RChandler234: can you also add this prop to the Select so the dropdown is aligned on the against the right side of the input box: 
```
              MenuProps={{
                anchorOrigin: {
                  vertical: 'bottom',
                  horizontal: 'right'
                },
                transformOrigin: {
                  vertical: 'top',
                  horizontal: 'right'
                }
              }}
```

RChandler234 (changes requested):

src/frontend/src/components/ChangeRequestDropdown.tsx
callmejim1226: I also think in that case, if B's submitter is the same as the current user and A is not, returning -1 would cause sorting the crs in reverse order where the current user's cr, B comes after cr, A `[A, B]`.

callmejim1226 (approved):
Everything looks good to me! Great job!

Reid has a good suggestion on component definition, so after revising it, I think this should be good to go.

RChandler234 (approved):

PR #1421, opened 2023-09-14
#1250: Try Catch and Toast uncaught Mutate Async

### LINKED ISSUE #1250: Frontend - Toast all mutate asyncs
### Description

When we make calls to our backend we use a mutate async to send the payload needed to the endpoint, these endpoints can all throw errors if something goes wrong and if they arent surrounded by a try catch they will crash the website which isnt pleasurable for the user, instead add the try catch and then toast the error that we get, there should only be like 3 or 4 mutate asyncs left to add this to

### Acceptance Criteria

no more uncaught mutate asyncs in codebase 

### Proposed Solution

use vscode search feature for mutateAsync and then check all the files that they are all using try catches and all the errors are being toasted to the user

### Mocks

epic #1255

sharonyang16:
## Changes

Surrounded all remaining uncaught mutate asyncs in a try catch. Toasted error messages for the mutate asyncs in CreateChangeRequest.tsx and DescriptionPageBlock.tsx, did not toast for the other three (will explain in Notes section).

## Notes

- Importing useToast and trying to toast error messages for auth.hooks.ts, CheckList.tsx, and ProposedSolutionsList.tsx would cause errors.
  - For auth.hooks.ts, this error would occur: 
auth hooks ts toast error
  - For CheckList.tsx and ProposedSolutionsList.tsx, they would fail their respective tests (the test results log is too long to include a screenshot of, but you can see them in the tests for the first two commits of this branch) and would give an error along the lines of "Cannot destructure property 'addToast' of 'useContext(...)' as it is null."

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #1250


tigerhippo (commented):
Nice job Sharon! I don't think you're doing anything wrong, I'm gonna have to ask Reid why the toast is either crashing the website or causing tests to fail. 

Also I found some hidden mutateAsyncs with ': someOtherFunctionName' next to them. This means that the mutateAsyncs have aliases and are used with those other names, and are not called using 'mutateAsync'. CreateReimbursementRequest file has 2 mutate asyncs that need a try/catch, EditReimbursementRequest file has 2 mutate asyncs without a try/catch, and SubmitToSaboModal file has 1 mutate async without a try/catch. I'll get back to you about how to get toast to work and then you can fix these.

src/frontend/src/pages/FinancePage/EditReimbursementRequest/EditReimbursementRequest.tsx
```diff
@@ -28,11 +30,19 @@ const EditReimbursementRequestPage: React.FC = () => {
 
   const onSubmit = async (data: ReimbursementRequestDataSubmission): Promise<string> => {
```
RChandler234, line 31: this is actually already try, catched where onSubmit is called so we don't need to do it here as well

src/frontend/src/pages/FinancePage/CreateReimbursementRequest.tsx
```diff
@@ -9,21 +9,30 @@ import { routes } from '../../utils/routes';
...
   if (createReimbursementRequestIsLoading || receiptsIsLoading) return <LoadingIndicator />;
 
   const onSubmit = async (data: ReimbursementRequestDataSubmission): Promise<string> => {
-    const { reimbursementRequestId } = await createReimbursementRequest(data);
-    await uploadReceipts({
-      id: reimbursementRequestId,
-      files: data.receiptFiles.map((file) => file.file!)
-    });
-    return reimbursementRequestId;
+    try {
```
RChandler234, line 23: same here, this is actually already try, catched where onSubmit is called so we don't need to do it here as well

src/frontend/src/pages/CreateChangeRequestPage/CreateChangeRequest.tsx
```diff
@@ -45,28 +45,34 @@ const CreateChangeRequest: React.FC<CreateChangeRequestProps> = () => {
...
         }
+      });
+    } catch (e) {
+      if (e instanceof Error) {
+        toast.error(e.message);
       }
-    });
+    }
 
     history.push(routes.CHANGE_REQUESTS);
```
RChandler234, line 77: this should go inside the try, also if have a try/catch inside a try/catch you can j delete the inner try/catch

src/frontend/src/hooks/auth.hooks.ts
```diff
@@ -16,16 +16,24 @@ export const useProvideAuth = () => {
...
   const signin = async (id_token: string) => {
-    const user = await mutateAsync(id_token);
-    setUser(user);
-    return user;
+    try {
+      const user = await mutateAsync(id_token);
+      setUser(user);
+      return user;
+    } catch (e) {
+      // do nothing; toasting error message will cause errors
```
RChandler234, line 35: this is the only file where toasting will actually cause an error, you should be able to do it in the other two places

src/frontend/src/components/CheckList.tsx
```diff
@@ -41,7 +41,11 @@ const CheckList: React.FC<CheckListProps> = ({ title, items, isDisabled }) => {
   };
 
   const handleCheck = async (idx: number) => {
-    await mutateAsync({ userId: auth.user!.userId, descriptionId: items[idx].id });
+    try {
+      await mutateAsync({ userId: auth.user!.userId, descriptionId: items[idx].id });
+    } catch (e) {
+      // do nothing; toasting error message will cause errors
```
RChandler234, line 47: you should be able to toast here without it causing an error

src/frontend/src/pages/ChangeRequestDetailPage/ProposedSolutionsList.tsx
```diff
@@ -33,16 +33,19 @@ const ProposedSolutionsList: React.FC<ProposedSolutionsListProps> = ({ proposedS
...
+      await mutateAsync({
+        submitterId: userId,
+        crId,
+        description,
+        scopeImpact,
+        timelineImpact,
+        budgetImpact
+      });
+    } catch (e) {
+      // do nothing; toasting error message will cause errors
```
RChandler234, line 47: you should be able to toast here without causing an error

src/frontend/src/hooks/auth.hooks.ts
RChandler234: so for this file just don't do the try, catch since we can't toast an error

RChandler234 (changes requested):

src/frontend/src/pages/ChangeRequestDetailPage/ProposedSolutionsList.tsx
```diff
@@ -24,6 +25,7 @@ const ProposedSolutionsList: React.FC<ProposedSolutionsListProps> = ({ proposedS
   const [showEditableForm, setShowEditableForm] = useState<boolean>(false);
   const auth = useAuth();
   const { isLoading, isError, error, mutateAsync } = useCreateProposeSolution();
+  const toast = useToast();
```
RChandler234, line 28: the tests for these two files fail bc this useToast() hook relies on a context that we wrap the whole app in called `<ToastProvider> </ToastProvider>`. This context works by basically adding a toast next to whatever content is a child of the toast provider (in this case our whole app). However, in the tests, we render components individually, so the component isn't wrapped in this context provider (hence the error we're getting). So, you need to wrap the component inside the renderComponent methods in both of these tests with a ToastProvider component to fix the problem

RChandler234 (changes requested):

src/frontend/src/hooks/auth.hooks.ts
```diff
@@ -16,10 +16,14 @@ export const useProvideAuth = () => {
   const [user, setUser] = useState<AuthenticatedUser | undefined>(undefined);
 
   const devSignin = async (userId: number) => {
-    const user = await mutateAsyncDev(userId);
-    setUser(user);
-    localStorage.setItem('devUserId', userId.toString());
-    return user;
+    try {
```
RChandler234, line 19: just remove the try, catch, since it's not doing anything

RChandler234 (changes requested):

src/frontend/src/hooks/auth.hooks.ts
sharonyang16: Ah, thanks! I had to revert some changes locally and lost my work and must've missed fixing that when I went back to work on it aughh

RChandler234 (approved):
LGTM 🔥

PR #1422, opened 2023-09-15
#716 jake wu chen buttonchange

### LINKED ISSUE #716: [Frontend] - Change ProposedSolutionForm to use NERSuccessButton
### Description

In the pursuit of making the looking and feel of our app consistent, we wanna use new components that have been made (NERSuccessButton/NERFailButton) in every place where we currently have a cancel/save button combo. By doing this, we not only make all these save/cancel buttons adhere to the new look, but we also allow ourselves to change how all these save buttons look in the future by changing only one file!

For this ticket, change the add buttons in the ProposedSolutionForm component to use the NERSuccessButton component rather than just the plain MUI button.

Put the cancel on the left, and the submit on the right

### Acceptance Criteria

The add button on ProposedSolutionForm use NERSuccessButton and not Button
The save/cancel buttons on that page looks like the new success/cancel buttons (green and red contained, see the edit members section on a teams specific page to see a good example of how it looks).
Verify nothing looks weird/ everything works as it did before.
Buttons should be 2 px apart, good example: https://github.com/Northeastern-Electric-Racing/FinishLine/pull/731

### Proposed Solution

All the props should be the same, all you need to do is swap out the old Button for the new NERSuccessButton

### Mocks

_No response_

jakewc12:
## Changes
Changed Add button from type Button to type NERSuccessButton and removed button import
image

Closes #716

### DIFF
--- a/src/frontend/src/pages/ChangeRequestDetailPage/ProposedSolutionForm.tsx
+++ b/src/frontend/src/pages/ChangeRequestDetailPage/ProposedSolutionForm.tsx
@@ -5,6 +5,7 @@
 
 import { Box, Button, Dialog, DialogContent, DialogTitle } from '@mui/material';
 import CloseIcon from '@mui/icons-material/Close';
+import NERSuccessButton from '../../components/NERSuccessButton';
 import * as yup from 'yup';
 import { Controller, useForm } from 'react-hook-form';
 import { yupResolver } from '@hookform/resolvers/yup';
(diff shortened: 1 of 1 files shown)


src/frontend/src/pages/ChangeRequestDetailPage/ProposedSolutionForm.tsx
```diff
@@ -5,6 +5,7 @@
 
 import { Box, Button, Dialog, DialogContent, DialogTitle } from '@mui/material';
```
walker-sean, line 6: Remove Button from here

walker-sean (changes requested):
Nice work! The linting check is failing because you have an unused import. After you fix that you'll be good to go!

walker-sean (changes requested):
Sweet, code looks great! One more thing I forgot to mention earlier, make sure to leave the relevant sections of the PR section up, so for this one, can you add back the checklist? Also, could you add a screenshot of the UI just to make sure everything looks right?

PR #1561, opened 2023-09-28
#1532 added seed data to orioles

### LINKED ISSUE #1532: [Seed Data] - Add at Least 8 Members to the Orioles Team
Epic: #976 

Quite frankly, the seed data for the local version of Finishline sucks right now. It has quite a small data size when compared to the production application, so we need to add a lot more data to it for it to be useful. The hope is that with these tickets, you can learn more about how prisma and our backend is setup, and we can get a more robust set of data to test the app locally with.

This ticket is to add at least 8 members (more if you want) to the Orioles Team. It would be cool if they're Orioles related, but it's fine if they're not. Please make an effort to make the names interesting (don't just put a bunch of bobs or johns)

To do this, you'll need to create the user data in users.seed.ts for each user, export them like the other users are, and then create the user in the database in the seed.ts file. After you've create the users, you can add them to the specified team under the `  /** Setting Team Members */` section in the seed.ts file

ScarlettBlaize:
## Changes

Added 9 more "people" to the users.seed.ts file and then added those people to the orioles team in the seed.ts file

Closes #1532

### DIFF
--- a/src/backend/src/prisma/seed-data/users.seed.ts
+++ b/src/backend/src/prisma/seed-data/users.seed.ts
@@ -244,6 +244,78 @@ const reidChandler: Prisma.UserCreateInput = {
   role: Role.HEAD
 };
 
+const brooksRobinson: Prisma.UserCreateInput = {
+  firstName: 'Brooks',
+  lastName: 'Robinson',
+  email: 'the.brooksters@orioles.com',
+  googleAuthId: 'fries',
+  role: Role.ADMIN
+};
+
+const jimPalmer: Prisma.UserCreateInput = {
+  firstName: 'Jim',
+  lastName: 'Palmer',
+  email: 'jimmy.palms@orioles.com',
+  googleAuthId: 'burger',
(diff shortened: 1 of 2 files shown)


ninest (approved):

src/backend/src/prisma/seed-data/users.seed.ts
```diff
@@ -244,6 +244,78 @@ const reidChandler: Prisma.UserCreateInput = {
   role: Role.HEAD
 };
 
+const brooksRobinson: Prisma.UserCreateInput = {
+  firstName: 'Brooks',
+  lastName: 'Robinson',
+  email: 'the.brooksters@orioles.com',
+  googleAuthId: 'fries',
+  role: Role.ADMIN
```
RChandler234, line 252: members on a team should generally be at minimum a member and at max a head

RChandler234 (changes requested):

RChandler234 (approved):
LGTM 🔥

PR #1617, opened 2023-10-08
#1516 fixed team member display

### LINKED ISSUE #1516: Fix Team Members Display
### Description

This should wrap around instead of getting cutoff like this.

### Acceptance Criteria

It looks good

### Proposed Solution

You will probably need to add a bunch of users to a team to correctly test this

### Mocks

image

kyang04:
## Changes

Fixed the team member display so that it wraps around the text area when the screen gets minimized

## Screenshots

image
image

Closes #1516


src/frontend/src/pages/TeamsPage/TeamMembersPageBlock.tsx
```diff
@@ -149,7 +149,7 @@ const TeamMembersPageBlock: React.FC<TeamMembersPageBlockProps> = ({ team }) =>
 
   const NonEditingMembersView = () => (
     <Grid container>
-      <Grid item xs={11} lg="auto">
+      <Grid item xs={11} lg="auto" style={{ maxWidth: 'calc(100% - 48px)' }}>
```
callmejim1226, line 152: Good job on adding `maxWidth` style here to wrap the members list. 

In fact, I think we can use the full width of this Grid to list members list since this Grid has limited width size already, as it's in `PageBlock` component as a child element.

In other words, I believe we can use `fit-content` instead of `calc(100% - 48px)` to wrap the text and avoid hard-coding the size value.

callmejim1226 (changes requested):
Great work! 👍 

Just left a minor comment to avoid hard-coding the size value.

src/frontend/src/pages/TeamsPage/TeamMembersPageBlock.tsx
callmejim1226: Alternatively, I think setting the max width as `100%` should also work, but let's ask it to Reid after we test out for `fit-content`.

callmejim1226 (approved):
Good job! LGTM!

RChandler234 (approved):
LGTM 🔥

PR #1667, opened 2023-10-25
#1529: Added more data to seed.ts and initialized them on users.seeds.ts

### LINKED ISSUE #1529: [Seed Data] - Add at Least 8 Members to the Justice League Team
Epic: #976 

Quite frankly, the seed data for the local version of Finishline sucks right now. It has quite a small data size when compared to the production application, so we need to add a lot more data to it for it to be useful. The hope is that with these tickets, you can learn more about how prisma and our backend is setup, and we can get a more robust set of data to test the app locally with.

This ticket is to add at least 5 members (more if you want) to the Justice League Team. It would be cool if they're Justice League related, but it's fine if they're not. Please make an effort to make the names interesting (don't just put a bunch of bobs or johns)

To do this, you'll need to create the user data in users.seed.ts for each user, export them like the other users are, and then create the user in the database in the seed.ts file. After you've create the users, you can add them to the specified team under the `  /** Setting Team Members */` section in the seed.ts file

patriots1:
## Changes

I added more justice league character seed data to seed.ts and initialized them on users.seed.ts. 

## Test Cases

I ran yarn test:frontend and yarn test:backend and saw the green pass symbol. I didn't create any new tests for my edits. 
- ...

Closes #1529

### DIFF
--- a/src/backend/src/prisma/seed-data/users.seed.ts
+++ b/src/backend/src/prisma/seed-data/users.seed.ts
@@ -88,7 +88,7 @@ const hawkgirl: Prisma.UserCreateInput = {
 const elongatedMan: Prisma.UserCreateInput = {
   firstName: 'Randolph',
   lastName: 'Dibney',
-  email: 'elongatedman@justiceleague.com',
+  email: 'elongatedmangit @justiceleague.com',
   googleAuthId: 'joigiug',
   role: Role.MEMBER
 };
--- a/src/backend/src/prisma/seed-data/users.seed.ts
+++ b/src/backend/src/prisma/seed-data/users.seed.ts
@@ -69,6 +69,70 @@ const wonderwoman: Prisma.UserCreateInput = {
   googleAuthId: 'wonderwoman'
 };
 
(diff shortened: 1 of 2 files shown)


ninest (approved):

RChandler234 (changes requested):
you need to write 'closes #ticket-number' at the end of the ticket description and make sure it's correctly linked on the right side under development

RChandler234 (changes requested):
also make sure yarn prisma:reset runs correctly and actually changes the seed data as well, it fails when I run it right now

patriots1:
Hi. I fixed the problem and successfully ran yarn prisma:reset, yarn test:frontend and yarn test:backend. I also fixed the issue regarding the ticket description.

RChandler234 (changes requested):
you need to add the created users to the justice league team as well

patriots1:
Hi. I added the members to the justice league team.

ninest (approved):

RChandler234 (approved):
LGTM 🔥

PR #1685, opened 2023-11-03
#1576 Threw an exception when there is an existing phone number in th…

### LINKED ISSUE #1576: Finance Fixes - Check for duplicate phone numbers before prisma call in backend (User Secure Settings)
### Description

Right now, in the backend, we have the unique specification on the phone number attribute of the user secure settings, but we don't check whether a phone number already exists before trying to set a user's secure settings. This results in a not very useful prisma error being thrown if someone tries to create a UserSecureSetting with a phone number that already exists. Therefore, we should add a check to the setUserSecureSettings Endpoint in the backend that throws a more useful 400 error if the payload contains a phone number that already exists in the database.

### Acceptance Criteria

- [ ] The setUserSecurreSettings Endpoint throws a 400 error if you try to submit a phone number that is already set as someone's phone number in the database

AShyamkumar19:
Threw an exception when there is an existing phone number in the user database and added 1 test

## Changes

- Added a feature where it checks for any duplication of phone number for every user

## Test Cases

- Tested on test file 

Closes #1576

### DIFF
--- a/src/backend/src/services/users.services.ts
+++ b/src/backend/src/services/users.services.ts
@@ -289,6 +289,15 @@ export default class UsersService {
     zipcode: string,
     phoneNumber: string
   ): Promise<string> {
+
+    const existingUser = await prisma.user_Secure_Settings.findFirst({
+      where: { phoneNumber }
+    });
+
+    if (existingUser) {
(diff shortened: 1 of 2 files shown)


walker-sean (approved):
LGTM

PR #1699, opened 2023-11-09
#1659 visual indication when tabbing through delete and add

### LINKED ISSUE #1659: Finance Fixes - You Should Be Able to Tab Through Product Section
### Description

Right now you can tab through all the buttons on the product table in a reimbursement request create/edit view, but the delete button and add buttons don't give any indication when they're selected. Ideally they should have the same slightly different look they have when you hover over them.

### Acceptance Criteria

- [ ] When you tab to a delete button, it has the same look as when you hover over it
- [ ] When you tab to an add product button, it has the same look as when you hover over it

### Proposed Solution
Look at the MUI docs to see how to change the look of something while focused on (which I believe are state these buttons are in when you tab to them)

msereeyothin:
## Changes

There is now a visual indication when tabbing through delete and add button of the products of the reimbursement request. 

## Screenshots

add product tabbed
delete button tabbed

Closes #1659

### DIFF
--- a/src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
+++ b/src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
@@ -25,6 +25,7 @@ import { ReimbursementProductCreateArgs, validateWBS, wbsPipe } from 'shared';
 import { Add, Delete } from '@mui/icons-material';
 import { Control, Controller, FieldErrors, UseFormSetValue } from 'react-hook-form';
 import { ReimbursementRequestFormInput } from './ReimbursementRequestForm';
+import { useTheme } from '@mui/system';
 
 interface ReimbursementProductTableProps {
   reimbursementProducts: ReimbursementProductCreateArgs[];
@@ -73,6 +74,9 @@ const ReimbursementProductTable: React.FC<ReimbursementProductTableProps> = ({
   const onCostBlurHandler = (value: number, index: number) => {
     setValue(`reimbursementProducts.${index}.cost`, parseFloat(value.toFixed(2)));
   };
+
+  const userTheme = useTheme();
+  const hoverColor = userTheme.palette.action.hover;
 
   return (
     <TableContainer>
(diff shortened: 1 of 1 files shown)


callmejim1226 (changes requested):
Great start!

When I ran the local Finishline app in dark mode, I checked the 'Add Product' button to get a white background after clicking it. 

Can we fix this and make sure the button looks good in the dark mode too?

image

msereeyothin:
It should now work in light and dark mode. However, the background color while hovering of the 'add product' button is slightly different to match the color of the focus.

src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
```diff
@@ -162,7 +166,10 @@ const ReimbursementProductTable: React.FC<ReimbursementProductTableProps> = ({
                     sx={{
                       margin: '4px',
                       '&:focus': {
-                        backgroundColor: '#fdf8f8'
+                        backgroundColor: hoverColor
```
callmejim1226, line 169: I think we can remove `&focus` prop from `sx` prop to remove the different color background on 'add product' button. It got removed in my local app after I removed `&focus` prop.

msereeyothin: Removing `&:focus` from the sx prop makes it so there is no indication when we tab through to the add product button. I think that would go against the acceptance criteria for this ticket?

callmejim1226: I see. According to the ticket description, your point sounds right. However, I don't get the point of us having a highlight show up on the button after we click on `ADD PRODUCT` and `delete icon`.

@RChandler234 Let us know if this is what you expected for this ticket.

RChandler234: you can make the onclick method for add product and it'll fix both problems:
```
(e) => {
                      appendProduct({
                        wbsNum: validateWBS(key),
                        name: '',
                        cost: 0
                      });
                      e.currentTarget.blur();
                    }
```

RChandler234 (changes requested):

callmejim1226 (approved):
LGTM! Great job!

RChandler234 (approved):
LGTM 🔥

PR #1733, opened 2023-12-05
#1605 Archive Team Endpoint

### LINKED ISSUE #1605: Archivable Teams - archive team endpoint
### Description

We need the ability to tell the database that a team has been archived or unarchived. create an endpoint for this function. 

### Acceptance Criteria

- [ ] new delete route teams/:id/archive
- [ ] Controller to get the current user and the team id out of the params
- [ ] Service function that marks the passed in team with the userId of the current user for the userDeleted field and the current date for the date deleted field if the team does not have a field for this already, if it does, mark these fields as null (unarchving it)
- [ ] Validate the user is an admin or above
- [ ] A team should not be archivable if it has any active projects, or incomplete projects. 
- [ ] Unit Tests

### Proposed Solution

View other endpoints for guidance

epic #1603 

### Mocks

_No response_

Aaryan1203:
## Changes

Created an endpoint to archive a team

## Notes

- made some changes to the teams transformer and team-types because of the new migration that added new fields
- Also added a project to one of the teams in the teams test data for my test to ensure that a team does not have any incomplete or active projects

## Test Cases

- on the tests file

Closes #1605


src/backend/src/services/teams.services.ts
```diff
@@ -284,4 +284,27 @@ export default class TeamsService {
...
+    if (team.projects.some((project) => project.wbsElement.status !== WBS_Element_Status.COMPLETE)) {
+      throw new HttpException(400, 'A team is not archivable if it has any active projects, or incomplete projects');
+    }
+
+    team.userArchivedId = user.userId;
+
+    team.dateArchived = team.dateArchived !== null ? new Date() : null;
+
+    return teamTransformer(team);
+  }
```
ninest, line 309: It seems like you are editing the variable team, but really updating it. To update the database, you will have to call prisma.team.update(...)

ninest (changes requested):

src/backend/tests/teams.test.ts
```diff
@@ -326,4 +326,50 @@ describe('Teams', () => {
...
+      expect(prisma.team.update).toHaveBeenCalledWith({
+        where: { teamId },
+        ...teamQueryArgs,
+        data: {
+          userArchivedId: superman.userId,
+          dateArchived: justiceLeague.dateArchived !== null ? new Date() : null
+        }
+      });
+
+      expect(res.userArchivedId).toBe(justiceLeagueUpdates.userArchivedId);
```
RChandler234, line 371: the mock call above passes, so you don't need to check the response in line 371/2. It's not really testing anything more, so it's not worth trying to mess with the janky mock stuff

src/backend/tests/teams.test.ts
```diff
@@ -354,7 +354,6 @@ describe('Teams', () => {
       const justiceLeagueUpdates = { ...justiceLeague, userArchivedId: 2, dateArchived: new Date() };
 
       vi.spyOn(prisma.team, 'update').mockResolvedValue(justiceLeagueUpdates);
-      const res = await TeamsService.archiveTeam(superman, justiceLeague.teamId);
```
RChandler234: you should keep this line, so it actually tests the service method/ passes the test, just remove the `const res` so that the linting check also passes

RChandler234 (changes requested):

src/shared/src/types/team-types.ts
```diff
@@ -15,6 +15,8 @@ export interface Team {
   members: User[];
   projects: ProjectPreview[];
   leads: User[];
+  userArchivedId?: number;
```
RChandler234, line 18: this should be userArchived

src/backend/src/transformers/teams.transformer.ts
```diff
@@ -19,7 +19,9 @@ const teamTransformer = (team: Prisma.TeamGetPayload<typeof teamQueryArgs>): Tea
       name: project.wbsElement.name,
       status: calculateProjectStatus(project)
     })),
-    leads: team.leads.map(userTransformer)
+    leads: team.leads.map(userTransformer),
+    userArchivedId: team.userArchivedId ?? undefined,
```
RChandler234, line 23: should be userArchived

src/backend/src/routes/teams.routes.ts
```diff
@@ -39,4 +39,6 @@ teamsRouter.post(
   TeamsController.createTeam
 );
 
+teamsRouter.delete('/:id/archive');
```
RChandler234, line 42: should be `:teamId`

RChandler234: I know the ticket said to make this a delete endpoint, but since we're even calling the endpoint archive instead of delete, I think it makes more sense to make it a post endpoint

src/backend/src/services/teams.services.ts
```diff
@@ -284,4 +284,38 @@ export default class TeamsService {
...
+   * @throws if the team is not found, the submitter has no privilege, the team has any projects that are not complete
+   */
+  static async archiveTeam(subimtter: User, teamId: string): Promise<Team> {
+    const team = await prisma.team.findFirst({
+      where: { teamId },
+      ...teamQueryArgs
+    });
+
+    if (!team) throw new NotFoundException('Team', teamId);
+
```
RChandler234, line 301: should also throw an error if it's already been deleted

src/backend/src/services/teams.services.ts
```diff
@@ -284,4 +284,38 @@ export default class TeamsService {
...
+    if (team.projects.some((project) => project.wbsElement.status !== WBS_Element_Status.COMPLETE)) {
+      throw new HttpException(400, 'A team is not archivable if it has any active projects, or incomplete projects');
+    }
+
+    const updatedTeam = await prisma.team.update({
+      where: { teamId },
+      ...teamQueryArgs,
+      data: {
+        userArchivedId: subimtter.userId,
+        dateArchived: team.dateArchived !== null ? new Date() : null
```
RChandler234, line 315: oh right, is this one where we archive/ unarchive with the same endpoint? if so you should put that in the comments somewhere

RChandler234 (changes requested):

src/backend/src/routes/teams.routes.ts
```diff
@@ -39,4 +39,6 @@ teamsRouter.post(
   TeamsController.createTeam
 );
```
walker-sean, line 41: Remove this space

src/backend/src/services/teams.services.ts
walker-sean: We hard delete teams so this should be covered by not finding the team

walker-sean (changes requested):
NIce work! One tiny nitpick

RChandler234 (approved):

PR #1856, opened 2024-01-20
#1781 Changed the quantity for material type to be a decimal

### LINKED ISSUE #1781: Quantity Should be a Double - Schema Change
### Description

We want 'Quantity' on a 'Material' in the schema to a 'Decimal' instead of an 'Int.' This is to facilitate a case when someone buys something like 20.5 ft of steel or smthn like that.

### Acceptance Criteria

- [ ] There's a new migration that converts quantity from an int to a decimal

### Proposed Solution

If you haven't done this before, you make the change to schema.prisma, then run yarn prisma:migrate, enter a migration name, then BOOM the migration will be created and applied.

You should test this locally to make sure that it still saves the old quantity data after running the migration

### Mocks

_No response_

Aaryan1203:
## Changes

Changed the schema for a Materials quantity from an int to a Decimal

## Test Cases

- Not too sure how to test it

## To Do

- Test

## Checklist

- [ ] No merge conflicts
- [ ] All checks passing

Closes #1781


src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/MaterialForm/MaterialFormView.tsx
```diff
@@ -184,7 +185,7 @@ const MaterialFormView: React.FC<MaterialFormViewProps> = ({
               <ReactHookTextField
                 name="quantity"
                 control={control}
-                errorMessage={errors.quantity}
+                errorMessage={errors.quantity?.root} // I just wrote root here for the ts test to pass but not too sure what should go here because I was getting a type error
```
RChandler234, line 188: it's because you set quantity to a Decimal, decimal might not be compatible with React-hook-forms. We might wanna keep it as a number (bc/ I believe that still keeps info about the decimal places) while it's in the form, then convert it to a Decimal to send to the backend

RChandler234 (changes requested):
I'm also getting a bug when editing BOMs rn where it says pdmFileName is a bad value. It might be because it's sending an empty string rather than not sending any value for pdmFileName

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/MaterialForm/MaterialForm.tsx
```diff
@@ -119,8 +131,9 @@ const MaterialForm: React.FC<MaterialFormProps> = ({ submitText, onSubmit, defau
 
   const onSubmitWrapper = (data: MaterialFormInput): void => {
     const price = Math.round(data.price * 100);
-    const subtotal = data.unitName ? price : data.quantity * price;
-    onSubmit({ ...data, subtotal: subtotal, price: price });
+    const quantity = Number(data.quantity);
+    const subtotal = data.unitName ? price : quantity * price;
```
RChandler234, line 135: I think I added this in, but I can't see any reason why it should be here. subtotal should just be quantity * price

src/backend/src/routes/projects.routes.ts
```diff
@@ -86,7 +86,7 @@ projectRouter.post(
   nonEmptyString(body('manufacturerName')),
   nonEmptyString(body('manufacturerPartNumber')),
   nonEmptyString(body('pdmFileName').optional()),
-  intMinZero(body('quantity')),
+  body('quantity'),
```
RChandler234, line 89: we should make a new validation method for decimals, see how we do it in validation.utils

RChandler234 (changes requested):

src/backend/src/routes/projects.routes.ts
Aaryan1203: What should it validate? like the fact that its a decimal greater than 0?

RChandler234: yup

RChandler234 (changes requested):
image
I think we need to do some rounding when calculating subtotal

RChandler234 (approved):
Great Job!

PR #1988, opened 2024-01-26
#1499: replaced instances of 'any' in AppContext.test.tsx

### LINKED ISSUE #1499: [Refactoring] - Remove `any` type from `src\frontend\src\tests\app\AppContext.test.tsx`
### Description

Remove the use of the `any` type from AppContext.test.tsx This may seem very simple, but the point of this is just to have you read through the codebase and understand the types/interfaces used in it, or make new ones that fit the shape of the data being used.

### Notes

### Acceptance Criteria

`any` is not used as a type in AppContext.test.tsx

vsp05:
## Changes

Changed props type from "any" to other values that made more sense in the file "AppContext.test.tsx".

Closes #1499

### DIFF
--- a/src/frontend/src/tests/app/AppContext.test.tsx
+++ b/src/frontend/src/tests/app/AppContext.test.tsx
@@ -9,9 +9,7 @@ import AppContext from '../../app/AppContext';
 vi.mock('../../app/AppContextQuery', () => {
   return {
     __esModule: true,
-    default: (props: {
-      children: string | number | boolean | {} | React.ReactNodeArray | React.ReactPortal | null | undefined;
-    }) => {
+    default: (props: { children: React.ReactNode | null }) => {
       return <div>app context query {props.children}</div>;
     }
   };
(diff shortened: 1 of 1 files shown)


src/frontend/src/tests/app/AppContext.test.tsx
```diff
@@ -9,7 +9,9 @@ import AppContext from '../../app/AppContext';
 vi.mock('../../app/AppContextQuery', () => {
   return {
     __esModule: true,
-    default: (props: any) => {
+    default: (props: {
+      children: string | number | boolean | {} | React.ReactNodeArray | React.ReactPortal | null | undefined;
```
jackturner106, line 13: Okay this is interesting! I like that you found a way to make it work, but I do think we should be able to tighten up the props a little bit. Could you try making the options limited to just React.ReactNode or possibly React.ReactNode | null and see if the tests still work?

jackturner106 (commented):
Good work! I think it should be possible to reduce the number of options there, I left a comment about that.

src/frontend/src/tests/app/AppContext.test.tsx
```diff
@@ -9,7 +9,7 @@ import AppContext from '../../app/AppContext';
 vi.mock('../../app/AppContextQuery', () => {
   return {
     __esModule: true,
-    default: (props: any) => {
+    default: (props: { children: React.ReactNode | null }) => {
```
RChandler234, line 12: you should be able to remove the | null

RChandler234 (changes requested):

jackturner106 (approved):
Looks good to me!

RChandler234 (approved):

PR #2050, opened 2024-02-02
#1518 Reply to Slack thread on CR approved/implemented

### LINKED ISSUE #1518: [Spike] - Slack Message in Thread when a CR is Approved/ Implemented
### Spike Type

Technical

### Goal

Figure out how to get the Slackbot to reply in a thread it has created

### Reason for Spike

Ideally the slackbot would reply in the original thread that it creates when a CR is created when the CR is approved (we could also move any reviewer requests into this thread as well), but we need to figure out how we would do this 

### Additional notes

_No response_

alaneng-neu:
## Changes

added new Slack utils to send messages as replies to other messages in a thread, sends replies to threads of the original CR notification in Slack with the new status (approved, implemented) and reacts to the message

## Screenshots

new column in table from schema migration
cr-new-schema

slack notification on CR approved
cr-approved-preview
cr-approved-thread

slack notification on CR denied
cr-denied-preview
cr-denied-thread

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #1518

### DIFF
--- a/src/backend/src/integrations/slack.ts
+++ b/src/backend/src/integrations/slack.ts
@@ -63,6 +63,7 @@ export const replyToMessageInThread = async (
   }
 };
 
+// to be changed or removed
 /**
  * Finds all messages with the search query from the slack bot
  * @param query - the query to search messages for
--- a/src/backend/src/integrations/slack.ts
+++ b/src/backend/src/integrations/slack.ts
@@ -1,4 +1,4 @@
-import { WebClient } from '@slack/web-api';
+import { ChatPostMessageResponse, WebClient } from '@slack/web-api';
 import { HttpException } from '../utils/errors.utils';
 
 const slack = new WebClient(process.env.SLACK_BOT_TOKEN);
@@ -9,6 +9,7 @@ const slack = new WebClient(process.env.SLACK_BOT_TOKEN);
  * @param message - the text content of the message being sent
  * @param link - the link for the button on the message
  * @param linkButtonText - the text for the button on the message
+ * @returns the channel id and timestamp of the created slack message
  */
 export const sendMessage = async (slackId: string, message: string, link?: string, linkButtonText?: string) => {
   const { SLACK_BOT_TOKEN } = process.env;
@@ -17,13 +18,15 @@ export const sendMessage = async (slackId: string, message: string, link?: strin
   const block = generateSlackTextBlock(message, link, linkButtonText);
 
   try {
-    await slack.chat.postMessage({
+    const response: ChatPostMessageResponse = await slack.chat.postMessage({
       token: SLACK_BOT_TOKEN,
       channel: slackId,
       text: message,
       blocks: [block],
       unfurl_links: false
     });
+
+    return response && response.channel && response.ts ? { channelId: response.channel, ts: response.ts } : null;
   } catch (error) {
     throw new HttpException(500, 'Error sending slack message, reason: ' + (error as any).data.error);
   }
@@ -60,31 +63,6 @@ export const replyToMessageInThread = async (
     });
   } catch (error) {
     throw new HttpException(500, 'Error sending slack reply to thread, reason: ' + (error as any).data.error);
-  }
-};
-
-// to be changed or removed
-/**
- * Finds all messages with the search query from the slack bot
- * @param query - the query to search messages for
- * @returns an array of matches with the channel id and timestamp of each match
- */
-export const findMessagesChannelIdTs = async (query: string) => {
-  const { SLACK_BOT_TOKEN } = process.env;
-  if (!SLACK_BOT_TOKEN) return;
-
-  try {
-    const result = await slack.search.messages({
-      token: SLACK_BOT_TOKEN,
-      query: `from:@finishline_by_ner ${query}`,
-      sort: 'timestamp',
-      sort_dir: 'asc'
-    });
-    if (result.ok && result.messages?.matches)
-      return result.messages.matches.map((match) => ({ channelId: match.channel?.id, ts: match.ts }));
-    return [];
-  } catch (error) {
-    throw new HttpException(500, 'Error searching slack messages, reason: ' + (error as any).data.error);
   }
 };
 
--- a/src/backend/src/integrations/slack.ts
+++ b/src/backend/src/integrations/slack.ts
@@ -26,7 +26,7 @@ export const sendMessage = async (slackId: string, message: string, link?: strin
       unfurl_links: false
     });
 
-    return response && response.channel && response.ts ? { channelId: response.channel, ts: response.ts } : null;
+    return response && response.channel && response.ts && { channelId: response.channel, ts: response.ts };
   } catch (error) {
     throw new HttpException(500, 'Error sending slack message, reason: ' + (error as any).data.error);
   }
--- a/src/backend/src/integrations/slack.ts
+++ b/src/backend/src/integrations/slack.ts
@@ -66,6 +66,28 @@ export const replyToMessageInThread = async (
 };
 
 /**
+ * Reacts to a slack message
+ * @param slackId - the channel id of the channel of the message to reply to
+ * @param parentTs - the timestamp of the message to reply to in a thread
+ * @param emoji - the emoji to react with
+ */
+export const reactToMessage = async (slackId: string, parentTs: string, emoji: string) => {
+  const { SLACK_BOT_TOKEN } = process.env;
+  if (!SLACK_BOT_TOKEN) return;
+
+  try {
+    await slack.reactions.add({
+      token: SLACK_BOT_TOKEN,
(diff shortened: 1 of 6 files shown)


src/backend/src/services/change-requests.services.ts
```diff
@@ -324,6 +326,18 @@ export default class ChangeRequestsService {
       }
     }
 
+    // send a reply to a CR's notifications of its updated status
+    try {
+      const relevantThreads = await prisma.message_Info.findMany({ where: { changeRequestId: foundCR.crId } });
```
RChandler234, line 331: this doesn't need to be in this try catch, you should pull it out. Also instead of doing it this way, you should do an if statement and only try to send the slack messages inside the if statement if relevant threads isn't undefined or an empty array. I know what you do here technically works, but this way is easier to read

src/backend/src/services/change-requests.services.ts
```diff
@@ -324,6 +326,18 @@ export default class ChangeRequestsService {
...
+      relevantThreads?.forEach(async (thread) => {
+        await sendSlackCRStatusToThread(thread.channelId, foundCR.crId, thread.timestamp, accepted);
```
RChandler234, line 333: move the try, catch into the helper method to send the message to thread that you made. If we're gonna do this, everywhere this is called to clean up the error message, then we should just abstract it away into the helper method

src/backend/src/services/change-requests.services.ts
```diff
@@ -324,6 +326,18 @@ export default class ChangeRequestsService {
       }
     }
 
+    // send a reply to a CR's notifications of its updated status
+    try {
+      const relevantThreads = await prisma.message_Info.findMany({ where: { changeRequestId: foundCR.crId } });
+      relevantThreads?.forEach(async (thread) => {
```
RChandler234, line 332: instead of this forEach, you should use map to create an array of promises, and resolve them with Promise.all (see how we use Promise.all elsewhere)

src/backend/src/integrations/slack.ts
```diff
@@ -9,49 +9,96 @@ const slack = new WebClient(process.env.SLACK_BOT_TOKEN);
...
-    await slack.chat.postMessage({
+    const response: ChatPostMessageResponse = await slack.chat.postMessage({
       token: SLACK_BOT_TOKEN,
       channel: slackId,
       text: message,
       blocks: [block],
       unfurl_links: false
     });
+
+    return response && response.channel && response.ts ? { channelId: response.channel, ts: response.ts } : null;
```
RChandler234, line 29: since you're returning null in the else case, you can just do && {your object} instead of ? {your object} : null

src/backend/src/integrations/slack.ts
```diff
@@ -9,49 +9,96 @@ const slack = new WebClient(process.env.SLACK_BOT_TOKEN);
...
+  const block = generateSlackTextBlock(message, link, linkButtonText);
+
+  try {
+    await slack.chat.postMessage({
+      token: SLACK_BOT_TOKEN,
+      channel: slackId,
+      thread_ts: parentTs,
+      text: message,
+      blocks: [block],
+      unfurl_links: false
```
RChandler234, line 62: What's this doing?

src/backend/src/utils/change-requests.utils.ts
```diff
@@ -27,33 +27,71 @@ export const sendSlackChangeRequestNotification = async (
...
   return Promise.all(msgs);
 };
 
+export const sendSlackCRStatusToThread = async (slackId: string, crId: number, ts: string, approved: boolean) => {
+  if (process.env.NODE_ENV !== 'production') return; // don't send msgs unless in prod
+  const msgs = [];
+  const fullMsg = `This Change Request was ${approved ? 'approved! :tada:' : 'denied.'} Click the link to view.`;
+  const fullLink = `https://finishlinebyner.com/cr/${crId}`;
+  const btnText = `View CR#${crId}`;
+  msgs.push(replyToMessageInThread(slackId, ts, fullMsg, fullLink, btnText));
```
RChandler234, line 69: so I think we can actually await the response and just return it like you did elsewhere. I think the msgs/ promise.all stuff was just copied over here when earlier code was written

src/backend/src/utils/change-requests.utils.ts
```diff
@@ -27,33 +27,71 @@ export const sendSlackChangeRequestNotification = async (
...
+};
+
+/**
+ * Adds the relevant slack notifications for a change request to the change request
+ *
+ * @param crId the change request to add the slack threads to
+ * @param notifications the slack threads to add to the change request
+ */
+export const addSlackThreadsToChangeRequest = async (crId: number, threads: { channelId: string; ts: string }[]) => {
+  threads.forEach(async (notification: { channelId: string; ts: string }) => {
```
RChandler234, line 81: do promise.all here as well

RChandler234 (changes requested):

src/backend/src/integrations/slack.ts
alaneng-neu: its a separate function for sending slack messages as a reply to a thread since in order for a message to be a reply, it requires a timestamp. if i added it as an optional arg to sendMessage, then it would either have to be before or after the link/linkButtonText params, meaning whichever of the args were the later half would have to coexist with the first arg, but all messages sent as a reply dont necessarily need to have a button, and vice versa. if you wanted to put that logic into sendMessage, then link/linkButtonText and ts would have to be consolidated into an optional options? param, which is an object with all three params.

src/backend/src/services/change-requests.services.ts
```diff
@@ -412,16 +426,24 @@ export default class ChangeRequestsService {
...
         const slackMsg =
           `${submitter.firstName} ${submitter.lastName} wants to activate ${createdCR.wbsElement.name}` +
           ` in ${createdCR.wbsElement.workPackage?.project.wbsElement.name}`;
-        await sendSlackChangeRequestNotification(team, slackMsg, createdCR.crId);
+        const sentNotifications: { channelId: string; ts: string }[] = await sendSlackChangeRequestNotification(
+          team,
+          slackMsg,
+          createdCR.crId
+        );
+        if (sentNotifications) notifications.push(...sentNotifications);
```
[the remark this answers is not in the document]
alaneng-neu: Not sure if this is possible because of the possibility of there being multiple teams meaning multiple notifications, where the notifications would be returned in different scopes since it uses teams.forEach, unless you want me to completely refactor `sendSlackChangeRequestNotification`

src/backend/src/integrations/slack.ts
RChandler234: sorry I meant specifically, unfurl_links: false

alaneng-neu: removed

src/backend/src/services/change-requests.services.ts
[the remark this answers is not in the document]
RChandler234: I see what you're saying, I think you good find a way to reduce a lot of this duplicate code would be by abstracting the code in the forEach call into a helper method, with inputs that allow it to be used in each of these 3 cases (cause I know we're sending 3 different messages)

RChandler234 (changes requested):

src/backend/src/utils/change-requests.utils.ts
```diff
@@ -27,33 +27,108 @@ export const sendSlackChangeRequestNotification = async (
...
-  const fullMsg = `:tada: Your Change Request was just reviewed! Clink the link to view! :tada:`;
+  const fullMsg = `:tada: Your Change Request was just reviewed! Click the link to view! :tada:`;
   const fullLink = `https://finishlinebyner.com/cr/${crId}`;
   const btnText = `View CR#${crId}`;
   msgs.push(sendMessage(slackId, fullMsg, fullLink, btnText));
 
   return Promise.all(msgs);
 };
 
+export const sendSlackCRStatusToThread = async (
```
RChandler234, line 79: add some documentation for this method cause it's got a good bit going on

src/backend/src/utils/change-requests.utils.ts
```diff
@@ -27,33 +27,108 @@ export const sendSlackChangeRequestNotification = async (
   message: string,
   crId: number,
   budgetImpact?: number
-) => {
-  if (process.env.NODE_ENV !== 'production') return; // don't send msgs unless in prod
-  const msgs = [];
+): Promise<{ channelId: string; ts: string }[]> => {
```
RChandler234, line 30: add some documentation to this method as well and make sure to describe the information that it returns

RChandler234 (changes requested):

RChandler234 (changes requested):

RChandler234 (approved):
LGTM 🔥

PR #2172, opened 2024-02-25
Updated projects services and test to hard delete material type

### LINKED ISSUE #2000: Delete Material Type Endpoint Should be Hard Delete
### Description

Right now we soft delete material types in the database (mark it as deleted by setting a dateDeleted on it), however, because we use name as the id, this would make it so that you could never make a material type with that name again once it's deleted. We should instead just have people hard delete the material type from the database and make a new one.

### Acceptance Criteria

- [ ] The delete material type service method should hard delete material types rather than setting a dateDeleted on it.
- [ ] Update any related tests

lphan48:
## Changes

Updated project.services.ts and project.tests.js to hard delete material type instead of setting dateDeteled.

## Test Cases

- does not work if user is not an admin or head
- does not work if material type does not exist
- does not work if the material type is already deleted
- works as expected

## Checklist

- [x]  All commits are tagged with the ticket number
- [x]  No linting errors / newline at end of file warnings
- [x]  All code follows repository-configured prettier formatting
- [x]  No merge conflicts
- [x]  All checks passing
- [x]  Screenshots of UI changes (see Screenshots section)
- [x]  Remove any non-applicable sections of this template
- [x]  Assign the PR to yourself
- [x]  No yarn.lock changes (unless dependencies have changed)
- [x]  Request reviewers & ping on Slack
- [x]  PR is linked to the ticket (fill in the closes line below)

Closes #2000

### DIFF
--- a/src/backend/src/controllers/projects.controllers.ts
+++ b/src/backend/src/controllers/projects.controllers.ts
@@ -312,8 +312,8 @@ export default class ProjectsController {
     try {
       const { materialTypeId } = req.params;
       const user = await getCurrentUser(res);
-      await ProjectsService.deleteMaterialType(materialTypeId, user);
-      return res.status(204).json({ message: `Successfully deleted material type with id ${materialTypeId}` });
+      const deletedMaterialType = await ProjectsService.deleteMaterialType(materialTypeId, user);
+      return res.status(200).json(deletedMaterialType);
     } catch (error: unknown) {
       next(error);
     }
--- a/src/backend/src/controllers/projects.controllers.ts
+++ b/src/backend/src/controllers/projects.controllers.ts
@@ -312,8 +312,8 @@ export default class ProjectsController {
     try {
       const { materialTypeId } = req.params;
       const user = await getCurrentUser(res);
-      const deletedMaterial = await ProjectsService.deleteMaterialType(materialTypeId, user);
-      res.status(200).json(deletedMaterial);
+      await ProjectsService.deleteMaterialType(materialTypeId, user);
+      return res.status(204).json({ message: `Successfully deleted material type with id ${materialTypeId}` });
     } catch (error: unknown) {
       next(error);
     }
--- a/src/backend/src/routes/work-packages.routes.ts
+++ b/src/backend/src/routes/work-packages.routes.ts
@@ -35,6 +35,21 @@ workPackagesRouter.post(
   WorkPackagesController.createWorkPackage
 );
 workPackagesRouter.post(
+  '/template/create',
+  nonEmptyString(body('templateName')),
+  nonEmptyString(body('templateNotes')),
+  nonEmptyString(body('workPackageName').optional()),
+  isWorkPackageStageOrNone(body('stage').optional()),
+  intMinZero(body('duration')),
+  body('expectedActivities').isArray(),
+  body('deliverables').isArray(),
+  body('blockedBy').isArray(),
+  nonEmptyString(body('blockedBy.*.name')),
+  isWorkPackageStageOrNone(body('blockedBy.*.stage').optional()),
+  validateInputs,
+  WorkPackagesController.createWorkPackageTemplate
+);
+workPackagesRouter.post(
   '/edit',
   intMinZero(body('workPackageId')),
   intMinZero(body('crId')),
--- a/src/backend/src/routes/work-packages.routes.ts
+++ b/src/backend/src/routes/work-packages.routes.ts
@@ -35,21 +35,6 @@ workPackagesRouter.post(
   WorkPackagesController.createWorkPackage
 );
 workPackagesRouter.post(
-  '/template/create',
-  nonEmptyString(body('templateName')),
-  nonEmptyString(body('templateNotes')),
-  nonEmptyString(body('workPackageName').optional()),
-  isWorkPackageStageOrNone(body('stage').optional()),
-  intMinZero(body('duration')),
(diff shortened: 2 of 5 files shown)


src/backend/tests/projects.test.ts
```diff
@@ -917,19 +917,13 @@ describe('Projects', () => {
...
-      );
-    });
-
     test('Delete Material Type works', async () => {
       vi.spyOn(prisma.material_Type, 'findUnique').mockResolvedValue(toolMaterial);
-      vi.spyOn(prisma.material_Type, 'update').mockResolvedValue({ ...toolMaterial, dateDeleted: new Date() });
-      const deletedMaterialType = await ProjectsService.deleteMaterialType('NERSoftwareTools', superman);
-      expect(deletedMaterialType.name).toBe('NERSoftwareTools');
-      expect(prisma.material_Type.update).toBeCalledTimes(1);
+      vi.spyOn(prisma.material_Type, 'update').mockResolvedValue(toolMaterial);
```
bderbs30, line 922: I think this should be delete instead of update

src/backend/tests/projects.test.ts
```diff
@@ -917,19 +917,13 @@ describe('Projects', () => {
       );
     });
 
-    test('Deleted Material Tye does not work if the material type is already deleted', async () => {
-      vi.spyOn(prisma.material_Type, 'findUnique').mockResolvedValue({ ...toolMaterial, dateDeleted: new Date() });
-      await expect(ProjectsService.deleteMaterialType('NERSoftwareTools', batman)).rejects.toThrow(
```
bderbs30, line 922: we want to do toolmaterials.name instead of the NERSoftwareTools. Do this in the other cases of this as well

src/backend/tests/projects.test.ts
```diff
@@ -917,19 +917,13 @@ describe('Projects', () => {
       );
```
bderbs30, line 917: In the notfound case you use NERSoftware Tools which does exist. instead use like 'fakeId' because this should not be return a not found exception

src/backend/tests/projects.test.ts
bderbs30: do this on like 909 as well

bderbs30 (approved):
LGTM

src/backend/src/services/projects.services.ts
```diff
@@ -949,18 +948,8 @@ export default class ProjectsService {
...
-      where: {
-        name: materialTypeId
-      },
-      data: {
-        dateDeleted: new Date()
-      }
-    });
-
-    return deletedMaterialType;
+    await prisma.material_Type.delete({ where: { name: materialTypeId } });
```
RChandler234, line 952: we still wanna return the deletedMaterialType in the response. This prisma call returns the object it deletes, so you can just return that to the controller and put it in the response

src/backend/src/controllers/projects.controllers.ts
```diff
@@ -312,8 +312,8 @@ export default class ProjectsController {
     try {
       const { materialTypeId } = req.params;
       const user = await getCurrentUser(res);
-      const deletedMaterial = await ProjectsService.deleteMaterialType(materialTypeId, user);
-      res.status(200).json(deletedMaterial);
+      await ProjectsService.deleteMaterialType(materialTypeId, user);
+      return res.status(204).json({ message: `Successfully deleted material type with id ${materialTypeId}` });
```
RChandler234, line 316: shouldn't need to return this, you can just do res.status

RChandler234 (changes requested):

src/backend/src/routes/work-packages.routes.ts
```diff
@@ -34,6 +34,21 @@ workPackagesRouter.post(
...
+  nonEmptyString(body('workPackageName').optional()),
+  isWorkPackageStageOrNone(body('stage').optional()),
+  intMinZero(body('duration')),
+  body('expectedActivities').isArray(),
+  body('deliverables').isArray(),
+  body('blockedBy').isArray(),
+  nonEmptyString(body('blockedBy.*.name')),
+  isWorkPackageStageOrNone(body('blockedBy.*.stage').optional()),
+  validateInputs,
+  WorkPackagesController.createWorkPackageTemplate
```
RChandler234, line 50: you can remove all this code since it's not related to hard deleting material type

RChandler234 (changes requested):

src/backend/src/services/projects.services.ts
```diff
@@ -949,17 +949,8 @@ export default class ProjectsService {
...
-    const deletedMaterialType = await prisma.material_Type.update({
-      where: {
-        name: materialTypeId
-      },
-      data: {
-        dateDeleted: new Date()
-      }
-    });
 
+    const deletedMaterialType = await prisma.material_Type.delete({ where: { name: materialTypeId } });
```
RChandler234, line 953: oops, should've mentioned this on the ticket, but we also need to make sure that the material type is not currently associated with any materials and throw an error if it is, because otherwise prisma will get mad at us.

material types should only really be deleted if they have a typo (in which case they shouldn't be associated with any materials)

RChandler234 (changes requested):

RChandler234 (changes requested):
Looks like you'll need to update the tests to account for the new error condition you added, lmk if you have questions ab that

RChandler234:
no longer doing

PR #2180, opened 2024-02-29
#1766 upgrade member to lead

### LINKED ISSUE #1766: Teams - Upgrading a Member to a Lead/Head automatically removes them as a member
### Description

Currently, you need to remove a member from a team, then add them as a head (or lead) if you want to upgrade them. This isn't a very intuitive user flow. The backend setTeamHead and setTeamLead endpoints should just automatically remove the user as a member from the project, then upgrade them to the desired level on the team.

### Acceptance Criteria

- [ ] The set team head endpoint automatically removes a user as a member (if they are) instead of throwing an error, then upgrades them to a head on that team
- [ ] The set team head endpoint automatically removes a user as a lead (if they are) instead of throwing an error, then upgrades them to a head on that team
- [ ] The set team lead endpoint automatically removes a user as a member (if they are) instead of throwing an error, then upgrades them to a lead on that team
- [ ] Ensure that all the downgrades also work automatically

### Proposed Solution

You should be able to just make an additional prisma call instead of throwing an error.

vsp05:
## Changes

Changed the endpoints for SetTeamHead, SetTeamLead, and SetTeamMembers that, when a member is given a new position, they are removed from their old position on that team.

## Notes:
When you try to set a team head that is already a team lead on that team, that team head won't appear in the dropdown menu for new team head options. I have changed the code in the backend that should make it work, but this code in the frontend is blocking that functionality from being available. This situation is the same for trying to demote a current team lead to a member of that team. Basically, I have fixed all of the backend endpoints so that it should work, but there are other pre-written blocks in the front-end that prevent it from being implemented. If you want me to change that in the front end too, I can. 

## Test Cases & Screenshots:
Case A:
Objective - When adding John Terry as a team lead for PlTeams, he is automatically removed as a team member.
Before photo:
Screenshot 2024-02-28 at 9 03 04 PM
After photo (I didn't manually delete John Terry as a member): 
Screenshot 2024-02-28 at 9 03 54 PM

Case B:
Objective - When Adding Paul Scholes as a team head for PlTeams, he is automatically removed as a team member.
Before Photo:
Screenshot 2024-02-28 at 9 04 53 PM
After photo:
Screenshot 2024-02-28 at 9 05 33 PM

Closes #1766


src/backend/src/services/teams.services.ts
```diff
@@ -73,9 +73,16 @@ export default class TeamsService {
 
     if (team.dateArchived) throw new HttpException(400, 'Cannot edit the members of an archived team');
 
-    if (team.leads.map((lead) => lead.userId).some((leadId) => userIds.includes(leadId)))
-      throw new HttpException(400, 'team leads cannot be members!');
-
+    // if the new members array includes a current lead on that team, that member will be deleted as a lead of that team
+    const newLeadsArr: number[] = [];
```
jackturner106, line 77: look into replacing all of this logic with the filter function, should be able to do 
`newLeadsArr = team.lead.filter((lead) => {some logic here about how userIds doesn't include lead.userId})`

src/backend/src/services/teams.services.ts
```diff
@@ -272,9 +298,18 @@ export default class TeamsService {
       throw new HttpException(400, 'A lead cannot be the head of the team!');
     }
 
-    if (team.members.map((member) => member.userId).some((memberId) => userIds.includes(memberId))) {
+    /*if (team.members.map((member) => member.userId).some((memberId) => userIds.includes(memberId))) {
```
jackturner106, line 301: Could you remove this commented out block

src/backend/tests/teams.test.ts
```diff
@@ -5,7 +5,7 @@ import { prismaTeam1, sharedTeam1, justiceLeague, primsaTeam2 } from './test-dat
 import teamQueryArgs from '../src/prisma-query-args/teams.query-args';
 import {
   alfred,
-  aquaman,
+  //aquaman,
```
jackturner106, line 8: if you're not using this could you remove it?

src/backend/tests/teams.test.ts
```diff
@@ -273,7 +273,8 @@ describe('Teams', () => {
       await expect(callSetTeamLeads).rejects.toThrow(expectedException);
     });
 
-    test('setTeamLeads lead is a member', async () => {
+    // This has now been changed in the backend so that a new team lead can be a current member
+    /*test('setTeamLeads lead is a member', async () => {
```
jackturner106, line 277: same with this block

jackturner106 (changes requested):
Good work getting this working! I think you should be able to replace most of the for loop logic with .filter, which would make the code generally more readable and maintainable. Other than that just removing the code blocks you commented out

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +159,20 @@ export default class TeamsService {
...
+    if (newHead && team.members.map((user) => user.userId).includes(newHead?.userId)) {
+      const newTeamArr: number[] = team.members
+        .filter((member) => member.userId !== newHead.userId)
+        .map((member) => member.userId);
+      this.setTeamMembers(submitter, teamId, newTeamArr);
+    }
+
+    // if the new head is a current leader on the team, they will be removed as a leader
+    // this will work in theory, but there is a block on the front-end that prevents the user from adding a current lead as the head of a team (they won't show up in the new head dropdown)
+    if (team.leads.map((lead) => lead.userId).includes(userId)) {
```
jackturner106, line 172: should this be .includes(newHead?.userId)

src/backend/src/services/teams.services.ts
```diff
@@ -73,9 +73,16 @@ export default class TeamsService {
 
     if (team.dateArchived) throw new HttpException(400, 'Cannot edit the members of an archived team');
 
-    if (team.leads.map((lead) => lead.userId).some((leadId) => userIds.includes(leadId)))
-      throw new HttpException(400, 'team leads cannot be members!');
-
+    // if the new members array includes a current lead on that team, that member will be deleted as a lead of that team
+    const newLeadsArr: number[] = [];
+    if (team.leads.map((lead) => lead.userId).some((leadId) => userIds.includes(leadId))) {
+      for (let i = 0; i < team.leads.length; i++) {
```
jackturner106, line 79: I like the other for loops being filters, could you do it for this one as well?

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +159,20 @@ export default class TeamsService {
 
     if (team.dateArchived) throw new HttpException(400, 'Cannot edit the head of an archived team');
 
-    if (newHead && team.members.map((user) => user.userId).includes(newHead?.userId))
-      throw new HttpException(400, 'Error: Team head cannot be a member');
+    // if the new head is a current member on the team, this will remove them as a member
+    if (newHead && team.members.map((user) => user.userId).includes(newHead?.userId)) {
```
jackturner106, line 163: Since you're checking that newHead exists, you should be able to drop the ? in the .includes (i.e. .includes(newHead.userId))

src/backend/src/services/teams.services.ts
jackturner106: and also do you need newHead &&?

src/backend/tests/teams.test.ts
```diff
@@ -273,18 +264,6 @@ describe('Teams', () => {
       await expect(callSetTeamLeads).rejects.toThrow(expectedException);
     });
 
-    test('setTeamLeads lead is a member', async () => {
```
jackturner106, line 276: Why did you remove this test? Could it be updated to just expect the correct response instead?

jackturner106 (changes requested):
I like it! I think you've got one more for loop, plus a few other small things

src/backend/src/services/teams.services.ts
vsp05: > should this be .includes(newHead?.userId)

If I'm not wrong, I think userId is the parameter that's passed into the function which represents the userId of the new Head, so saying newHead?.userId might be redundant? But if that's incorrect I'll definitely change that. 


> and also do you need newHead &&?

yeah I think I do, I just forgot that. I'll go ahead and fix that now.

src/backend/src/services/teams.services.ts
vsp05: > I like the other for loops being filters, could you do it for this one as well?

Yeah I think I must've missed this one. I'll fix that now as well.

src/backend/src/services/teams.services.ts
vsp05: > Since you're checking that newHead exists, you should be able to drop the ? in the .includes (i.e. .includes(newHead.userId))

Oh ok that makes sense, I'll fix that.

src/backend/tests/teams.test.ts
vsp05: Sorry, that's what I meant to do, but I think I must've accidentally removed it. I'll fix that real quick too.

vsp05: > Why did you remove this test? Could it be updated to just expect the correct response instead?

Oh wait, the reason that I removed this is because with my new changes, a new team lead can be a member of the team (at least, the backend code won't throw an error). Therefore, my thought process was that this test endpoint isn't necessary. If it is actually necessary, how would I go about writing that test function?

src/backend/src/services/teams.services.ts
jackturner106: As far as using newHead?, you use it on line 163, but don't use it on line 172, and I think they should be the same? If it isn't necessary, can it be removed from line 163?

src/backend/tests/teams.test.ts
jackturner106: Could you change it so that instead of checking if the endpoint throws an error when a member is set to head, it checks that the member was actually removed from the members list.

jackturner106: The way you could do that is by calling the setTeamHead endpoint and saving the result: `const res = await TeamsService.setTeamHead(...args);`, and then checking that in res the list of members doesn't include the new head

jackturner106 (changes requested):

src/backend/src/services/teams.services.ts
vsp05: Oh yeah good point, I'll remove it now.

src/backend/tests/teams.test.ts
vsp05: Ohh ok. Yeah I'll try to fix that right now

vsp05: Ok so I think I'm getting close, but my current test runs with an error that says "Cannot edit the leads of an archived team", even though the justiceLeague team isn't archived. Do you have an idea of why this could be? I'll attach some screenshots here. Thanks!

Mock justiceLeague team:
Screenshot 2024-03-24 at 10 05 51 PM

The error that is showing in the "test results" section:
Screenshot 2024-03-24 at 10 06 56 PM

My code for the actual test:
Screenshot 2024-03-24 at 10 06 20 PM

src/backend/tests/mocked/teams.test.ts
```diff
@@ -274,15 +274,13 @@ describe('Teams', () => {
...
-
-      const callSetTeamLeads = async () => await TeamsService.setTeamLeads(flash, sharedTeam1.teamId, [aquaman.userId]);
+      vi.spyOn(prisma.user, 'findUnique').mockResolvedValue(aquaman);
+      vi.spyOn(prisma.team, 'findUnique').mockResolvedValue(justiceLeague);
 
-      const expectedException = new HttpException(400, 'A lead cannot be a member of the team!');
+      const res = await TeamsService.setTeamLeads(flash, justiceLeague.teamId, [aquaman.userId]);
 
-      await expect(callSetTeamLeads).rejects.toThrow(expectedException);
+      //expect(res.leads.map((user) => user.userId)).toContain(aquaman.userId);
```
jackturner106, line 282: Can you uncomment this line so that the test actually checks that aquaman is now a lead

jackturner106 (changes requested):
looking good!

jackturner106 (changes requested):
This is good, could you also update the front end so that leads can be selected as the new head? In TeamMembersPageBlock.tsx you should be able to just remove the filter on memberOptions

jackturner106 (dismissed):
Looks good to me

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +156,20 @@ export default class TeamsService {
 
     if (team.dateArchived) throw new HttpException(400, 'Cannot edit the head of an archived team');
 
-    if (newHead && team.members.map((user) => user.userId).includes(newHead?.userId))
-      throw new HttpException(400, 'Error: Team head cannot be a member');
+    // if the new head is a current member on the team, this will remove them as a member
+    if (newHead && team.members.map((user) => user.userId).includes(userId)) {
+      const newTeamArr: number[] = team.members
+        .filter((member) => member.userId !== newHead.userId)
```
RChandler234, line 162: just use userId here instead of newHead.userId, since that's what you're doing in the if statement

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +156,20 @@ export default class TeamsService {
 
     if (team.dateArchived) throw new HttpException(400, 'Cannot edit the head of an archived team');
 
-    if (newHead && team.members.map((user) => user.userId).includes(newHead?.userId))
-      throw new HttpException(400, 'Error: Team head cannot be a member');
+    // if the new head is a current member on the team, this will remove them as a member
+    if (newHead && team.members.map((user) => user.userId).includes(userId)) {
+      const newTeamArr: number[] = team.members
```
RChandler234, line 161: newTeamMembersIds

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +156,20 @@ export default class TeamsService {
 
     if (team.dateArchived) throw new HttpException(400, 'Cannot edit the head of an archived team');
 
-    if (newHead && team.members.map((user) => user.userId).includes(newHead?.userId))
-      throw new HttpException(400, 'Error: Team head cannot be a member');
+    // if the new head is a current member on the team, this will remove them as a member
```
RChandler234, line 159: If the new head is a current member on the team, remove them as a member

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +156,20 @@ export default class TeamsService {
...
+    if (newHead && team.members.map((user) => user.userId).includes(userId)) {
+      const newTeamArr: number[] = team.members
+        .filter((member) => member.userId !== newHead.userId)
+        .map((member) => member.userId);
+      this.setTeamMembers(submitter, teamId, newTeamArr);
+    }
+
+    // if the new head is a current leader on the team, they will be removed as a leader
```
RChandler234, line 167: if the new head is a current lead on the team, remove them as a lead*

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +156,20 @@ export default class TeamsService {
...
+    // this will work in theory, but there is a block on the front-end that prevents the user from adding a current lead as the head of a team (they won't show up in the new head dropdown)
```
RChandler234, line 168: remove this comment and just change it in the frontend

src/backend/src/services/teams.services.ts
```diff
@@ -152,8 +156,20 @@ export default class TeamsService {
...
+    if (newHead && team.leads.map((lead) => lead.userId).includes(userId)) {
+      const newLeadsArr: number[] = team.leads.map((lead) => lead.userId).filter((leadId) => leadId !== userId);
```
RChandler234, line 170: newTeamLeadsIds*

src/backend/src/services/teams.services.ts
```diff
@@ -272,8 +288,12 @@ export default class TeamsService {
       throw new HttpException(400, 'A lead cannot be the head of the team!');
     }
 
+    // removes the new leads as current members of the given team (if they are current members of that team)
     if (team.members.map((member) => member.userId).some((memberId) => userIds.includes(memberId))) {
-      throw new HttpException(400, 'A lead cannot be a member of the team!');
+      const newMembersArr: number[] = team.members
```
RChandler234, line 293: newTeamMembersIds

src/backend/src/services/teams.services.ts
```diff
@@ -272,8 +288,12 @@ export default class TeamsService {
...
+        .map((member) => member.userId)
+        .filter((memberId) => !userIds.includes(memberId));
```
RChandler234, line 295: you should make a `removeUserFromTeam` helper method for this map and filter you're calling each time

src/backend/src/services/teams.services.ts
```diff
@@ -73,8 +73,12 @@ export default class TeamsService {
 
     if (team.dateArchived) throw new HttpException(400, 'Cannot edit the members of an archived team');
 
-    if (team.leads.map((lead) => lead.userId).some((leadId) => userIds.includes(leadId)))
-      throw new HttpException(400, 'team leads cannot be members!');
+    // if the new members array includes a current lead on that team, that member will be deleted as a lead of that team
+    // this should work in theory, but there is a block on the front-end that prevents the user from adding a current lead as a member of a team (they won't show up in the new member dropdown)
```
RChandler234, line 77: remove this comment and just change it in the frontend

src/backend/src/services/teams.services.ts
```diff
@@ -73,8 +73,12 @@ export default class TeamsService {
...
+    if (team.leads.map((lead) => lead.userId).some((leadId) => userIds.includes(leadId))) {
+      const newLeadsArr = team.leads.filter((lead) => !userIds.includes(lead.userId)).map((lead) => lead.userId);
```
RChandler234, line 79: newTeamLeadsIds

RChandler234 (changes requested):

src/backend/src/utils/teams.utils.ts
```diff
@@ -72,3 +72,8 @@ export const getTeamsFromUsers = (users: UserWithTeams[]): Team[][] => {
     return teams;
   });
 };
+
+export const removeUsersFromTeam = (currentUsers: User[], usersToRemove: User[]): number[] => {
```
RChandler234, line 76: this actually removes a list of users from another list of users, so I'd rename it to be more specific to that/ add a comment specifying what it's doing

RChandler234: love how clean this makes the code tho

src/backend/src/utils/users.utils.ts
```diff
@@ -50,6 +50,14 @@ export const getPrismaQueryUserIds = (users: User[]) => {
   return userIds;
 };
 
+export const getPrismaQueryFromIds = (userIds: number[]) => {
```
RChandler234, line 53: rename to getPrismaQueryFromUserIds

RChandler234 (changes requested):

src/backend/src/utils/teams.utils.ts
```diff
@@ -72,3 +72,15 @@ export const getTeamsFromUsers = (users: UserWithTeams[]): Team[][] => {
...
 };
+
+/**
+ * Removes all users in the second list from the first list. Returns the userIds of
+ * all users in the first list filtered to exclude those users in the second list.
+ * @param currentUsers The primary list of users
+ * @param usersToRemove the list of users to remove from currentUsers
+ * @returns the userIds of all users in currentUsers that aren't in usersToRemove
+ */
+export const removeUsersFromList = (currentUsers: User[], usersToRemove: User[]): number[] => {
```
walker-sean, line 83: You could make this return type of `User[]` to avoid having to define another user utils function like this:
```
const userIdsToRemove = usersToRemove.map((user) => user.userId);
return currentUsers.filter((user) => !userIdsToRemove.includes(user.userId))
```

walker-sean (changes requested):
Nice !

walker-sean (approved):

PR #2196, opened 2024-03-09
#issue-2003 added info icon to the manufacturer field

sajayi12:
## Changes

I have added the info icon tool tip to the manufacturer field for the BOM tab

## Screenshots
Screenshot 2024-03-09 at 2 12 34 AM

Screenshot 2024-03-09 at 2 12 14 AM

Screenshot 2024-03-11 at 10 05 47 PM

Screenshot 2024-03-11 at 10 19 14 PM

Screenshot 2024-03-11 at 11 30 00 PM

### DIFF
--- a/src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/MaterialForm/MaterialFormView.tsx
+++ b/src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/MaterialForm/MaterialFormView.tsx
@@ -124,7 +124,7 @@ const MaterialFormView: React.FC<MaterialFormViewProps> = ({
             <FormLabel>
               Manufacturer
               <DynamicTooltip title={`Make sure to enter the manufacturer and not the distributor!`}>
-                <InfoIcon sx={{ marginTop: '5px', marginLeft: 0.4, height: 11 }} />
+                <InfoIcon sx={{ marginTop: '5px', marginLeft: 0.2, height: 11 }} />
               </DynamicTooltip>
             </FormLabel>
             <Controller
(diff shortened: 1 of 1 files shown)


martin0he (changes requested):
the icon is too big, use the sx prop inside the infoicon to change the height of the Icon and a margin from the left to create some distance from the text and the icon

martin0he (changes requested):
icon is still too large. make it smaller, look at the text fields and how the manufacturer one is indented downwards (we don't want that). halve the height (or whatever looks better) and reduce the left padding to smth like .5 or something like that

martin0he (changes requested):
make it just a bit smaller and also add some vertical margin (marginTop) by a few pixels so that the bottom of the icon is aligned with the bottom of the text if that makes sense

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/MaterialForm/MaterialFormView.tsx
```diff
@@ -119,7 +121,12 @@ const MaterialFormView: React.FC<MaterialFormViewProps> = ({
         </Grid>
         <Grid item xs={4}>
           <FormControl fullWidth>
-            <FormLabel>Manufacturer</FormLabel>
+            <FormLabel>
+              Manufacturer
+              <DynamicTooltip title={`Make sure to enter the manufacturer and not the distributor!`}>
+                <InfoIcon sx={{ marginTop: '5px', marginLeft: 0.4, height: 11 }} />
```
martin0he, line 127: make margin left 0.2 and you're good

martin0he (changes requested):
halve the left margin again and also update your screenshots to only include the latest actual ones get rid of the old ones

martin0he (approved):
LGTM

Issue #2136, opened 2024-02-17
Design Review Calendar - Design Review Edit Page


RChandler234:
### Description

You should branch off of and merge into this feature branch: `feature/design-review`

When you click into a confirmed or scheduled design review, we want it to pop up the scheduling page below. This ticket is just to create the frontend portion of this ticket, hookup to the backend will happen in another ticket. The icons on the chart correspond to other scheduled design reviews (the icons are per team type - software, electrical, etc.). 

### Acceptance Criteria

- [x] The chart should show the combination of all the availabilities of all the required and optional attendees on the design review.
- [x] When you hover over a section in the chart, it should display the unavailable and available users on the right
- [x] The required and optional fields should be the same autocompletes we use to select members on the teams page
- [ ] If the current user's schedule settings are not set, the warning shown below should pop up with a link to the schedule settings page and the 'Finalize' button should be disabled
- [x] When you click finalize, the modal shown below should pop up, allowing the user to set (or edit any of the shown fields)
- [x] The modal should also show any conflicting design reviews so the scheduler can ensure that none of them will cause an issue
- [ ] It's not pictured in the mocks, but when you click on a section to schedule the meeting, it should highlight that box on the calendar and show the selected time on the right (might need to workshop this a bit depending on how tricky it is.

### Proposed Solution

Most of the functionality other than the availability chart we've implemented elsewhere, so you can look for examples. For the chart, you should coordinate with whoever's doing #2133 so they look similar.

### Mocks
Default View:
<img width="729" alt="image" src="https://github.com/Northeastern-Electric-Racing/FinishLine/assets/29521172/b687a972-5c0d-4b03-b877-f281c3d6535d">
When you click Finalize:
<img width="625" alt="image" src="https://github.com/Northeastern-Electric-Racing/FinishLine/assets/29521172/0a8b3d03-6dc2-43d5-b802-9cdc477b8462">

PR #2215, opened 2024-03-17
#2136 finalize design review modal

superhvarn:
## Changes
Aaryan - 
created a modal and linked it to the user availability page via finalize button
I also changed the look for the meeting type cuz i thought it looked cooler, lmk if I should change it to what it showed in the mock
## Screenshots

image

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes # (issue #2136 )


src/frontend/src/pages/CalendarPage/DesignReviewDetailPage/FinalizeDesignReviewModal.tsx
```diff
@@ -0,0 +1,56 @@
...
+  };
+
+  return (
+    <NERModal
+      open={open}
+      title={''}
+      onHide={handleClose} // Call handleClose function to close the modal
+      onSubmit={onSubmit}
+    >
+      <div>
```
Aaryan1203, line 21: We don't want to use divs or normal HTML tags, try using grids and custom components instead

Aaryan1203 (changes requested):
requested some changes

src/frontend/src/pages/CalendarPage/DesignReviewDetailPage/FinalizeDesignReviewModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+  setOpen: (val: boolean) => void;
+  onSubmit?: () => void;
+}
+
+const FinalizeDesignReviewModal = ({ open, setOpen, onSubmit }: FinalizeDesignReviewProps) => {
+  const handleClose = () => {
+    setOpen(false);
+  };
+
+  const names = ['jdel', 'dehkdh', 'dehjkdhk', 'dsahksdh', 'dsajkldsaj'];
```
RChandler234, line 27: this should be a prop that gets passed in. also that section of the modal should only show up if there are any conflicting meetings to show

src/frontend/src/pages/CalendarPage/DesignReviewDetailPage/FinalizeDesignReviewModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+            <ToggleButtonGroup color="primary" value={meetingType} onChange={handleMeetingTypeChange}>
+              <ToggleButton value="virtual">Virtual</ToggleButton>
+              <ToggleButton value="inPerson">In-person</ToggleButton>
+            </ToggleButtonGroup>
+          </Grid>
+          <Grid style={{ display: 'flex', alignItems: 'center', marginBottom: 50 }}>
+            <Typography style={{ textDecoration: 'underline', marginBottom: 10, fontSize: '1.2em', marginRight: 135 }}>
+              Location:
+            </Typography>
+            <FormControl style={{ backgroundColor: 'grey', width: '45%', borderRadius: 5 }} fullWidth>
```
RChandler234, line 71: no need to do the background color changes on any of these inputs, I know it's different than the mocks, but we don't really do this anywhere else on the site

src/frontend/src/pages/CalendarPage/DesignReviewDetailPage/FinalizeDesignReviewModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+  const handleMeetingTypeChange = (_event: any, newMeetingType: string[]) => {
+    setMeetingType(newMeetingType);
+  };
+
+  const handleLocationChange = (event: any) => {
+    setLocation(event.target.value);
+  };
+
+  return (
+    <NERModal open={open} title={''} onHide={handleClose} onSubmit={onSubmit} submitText="Finalize">
```
RChandler234, line 42: Modal should be titled 'Finalize Design Review Details'

RChandler234 (changes requested):
ok so, I've realized we need a way to actually see what the set time is/ set a new time on the design review detail page. Can you add that  (I think we can do name and a day/ time range selector on one row and then required and optional attendee selectors on the row below)

We should also be displaying that info (when the design review is actually scheduled for) at the top of the modal

walker-sean (changes requested):
Love how this is coming together!

Quick review, will look through code later: 
- We need to have a field for zoom link (preferably it is disabled, is cleared, and does not submit when the meeting is not online - and same thing with location/in person)
- The time selected must be on an hour - we could use a validator but it would be nicer to not let users select a time that doesn't fall on an hour.
- Display the name and when the dr is scheduled at the top of the modal

src/frontend/src/pages/CalendarPage/DesignReviewDetailPage/FinalizeDesignReviewDetailsModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+    setMeetingType(newMeetingType);
+  };
+
+  const handleLocationChange = (event: any) => {
+    setLocation(event.target.value);
+  };
+  return (
+    <NERModal
+      open={open}
+      title={`${designReviewName} on ${selectedDateTime?.toDateString()} at ${selectedDateTime?.toLocaleTimeString()}`}
```
walker-sean, line 52: This is showing even the seconds, could you convert it to something like 12PM

src/frontend/src/pages/CalendarPage/DesignReviewDetailPage/FinalizeDesignReviewDetailsModal.tsx
```diff
@@ -0,0 +1,139 @@
...
+                      width: '80%',
+                      height: '90px',
+                      overflowY: 'auto',
+                      backgroundColor: 'grey',
+                      padding: 1
+                    }}
+                  >
+                    {designReviews.map((designReview, index) => (
+                      <Typography key={index} style={{ color: 'black', borderTop: '1px solid black' }}>
+                        {designReview}
```
walker-sean, line 125: Also show the time of these

src/frontend/src/pages/CalendarPage/DesignReviewDetailPage/FinalizeDesignReviewModal.tsx
```diff
@@ -0,0 +1,108 @@
...
+              <ToggleButton value="inPerson">In-person</ToggleButton>
+            </ToggleButtonGroup>
+          </Grid>
+          <Grid style={{ display: 'flex', alignItems: 'center', marginBottom: 50 }}>
+            <Typography style={{ textDecoration: 'underline', marginBottom: 10, fontSize: '1.2em', marginRight: 135 }}>
+              Location:
+            </Typography>
+            <FormControl style={{ backgroundColor: 'grey', width: '45%', borderRadius: 5 }} fullWidth>
+              <InputLabel>Location</InputLabel>
+              <Select value={location} onChange={handleLocationChange} variant="outlined">
```
walker-sean, line 73: Location is just a string, so this should be a text input. This is for physical locations, so zoom isn't an option. Its usually gonna be in the bay so if we could do like "The Bay" or a custom input then we should, but if not we should just do a text input

walker-sean (changes requested):
Looking really good!

RChandler234:
Several Comments:

- None of the useEffect, useMemo stuff is necessary. While they are tools that React has, you should use them very sparingly because they cause re-renders (or limit them in useMemos case). For useMemo, I would almost never use it (see more details below). For useEffect, in our case we store all our state in the backend, so in most cases it makes the most sense to refetch new state from the backend using a query invalidation and cause re-renders using useState or props (whenever either changes it causes the re-render of a page).
- You should try whenever possible to localize super weird/ intensive data reorganizing. For instance you'll see I moved a lot of the existingMeetingData/ conflictingMeeting stuff down a component or two. This 1, helps clear up the main file you were working, and 2, helps reduce the load on re-rendering because the component has fewer sub-component that it will need to rerender, and 3 can help reduce the number of props you're passing around if you can move multiple calculations
- When you're passing around data, it's usually better to do name transformation only when you need to. You can see the way I passed around the selectedDate or designReview rather than the specific name because there's no need to limit what we're passing to just the designReview name, especially when we're going through multiple layers of components.
- I think the main failing of this PR was that things got way too complicated. You should whenever possible be making things as simple as possible (especially with big wire-ups like this) because, at least in my book, simple = better (easier for other people to use, less prone to error, shorter code). Ways to mitigate this are: 
- 1. trying to make sure components are properly abstracted (a component should have one purpose and anything not related to that should tried to be moved to a sub component or a helper method where possible)
- 2. I kinda like to say "don't do too much" (for example those time pickers are cool, but for our purposes are a bit complicated and give the user access to options they can't actually choose)
- 3. If you feel yourself getting to a point where you're like "this file is getting above 200 lines" or "wow this is super complicated," either take a bit and come back to the code or ask someone to take a look because (expect in rare cases) you're probably overcomplicating something

tl;dr I think you did really well on this PR, all of the logic was correct, I literally just reshuffled things to make the code nicer and it mostly works fine. The main takeaway is "[Keep it Simple Stupid](https://en.wikipedia.org/wiki/KISS_principle)"

RChandler234:
I'm still fixing up some the design/ the finalize functionality but this code should look a lot nicer

RChandler234 (approved):
LGTM 🔥

PR #2345, opened 2024-04-12
#2344 no more wbs proposed changes

### LINKED ISSUE #2344: Get rid of WbsProposedChanges
### Desired Changes

Instead of having WbsProposedChanges, we should only have ProjectProposedChanges and WorkPackackgeProposedChanges and have both of those include all the fields from workPackageProposedChanges. The route, controller and service functions all have to be updated to reflect this.

### Screenshots (as needed)

_No response_

Aaryan1203:
## Changes
Got rid of WbsProposedChanges

## Notes
Had to set default values for the wbs proposed changes, there has to be a better way to do that

##Todo
Fix the default value stuff

## Checklist

- [ ] No merge conflicts
- [ ] All checks passing

Closes #2344


src/backend/src/utils/validation.utils.ts
```diff
@@ -102,6 +98,13 @@ const workPackageProposedChangesExists = (validationObject: ValidationChain): Va
 
 export const workPackageProposedChangesValidators = [
   body('workPackageProposedChanges').optional(),
+  nonEmptyString(wbsProposedChangesExists(body('workPackageProposedChanges.name'))),
```
RChandler234, line 101: change to use workPackageProposedChangesExists

RChandler234: same to all below

src/backend/src/utils/validation.utils.ts
```diff
@@ -64,23 +64,19 @@ const wbsProposedChangesExists = (validationObject: ValidationChain): Validation
...
-  intMinZero(body('wbsProposedChanges.projectManagerId').optional())
-];
-
 const projectProposedChangesExists = (validationObject: ValidationChain): ValidationChain => {
   return validationObject.if((value: any, { req }: any) => req.body.projectProposedChanges);
 };
 
 export const projectProposedChangesValidators = [
   body('projectProposedChanges').optional(),
+  nonEmptyString(wbsProposedChangesExists(body('projectProposedChanges.name'))),
```
RChandler234, line 73: change to projectProposedChangesExists, same for all below

src/backend/src/services/change-requests.services.ts
```diff
@@ -614,87 +613,93 @@ export default class ChangeRequestsService {
...
-          if (!projectLead) throw new NotFoundException('User', projectLeadId);
-        }
+    if (projectProposedChanges && workPackageProposedChanges) {
+      throw new HttpException(400, "Change Request can't be on both a project and a work package");
+    } else if (!projectProposedChanges && !workPackageProposedChanges) {
+      throw new HttpException(
+        400,
+        'Change Request with proposed changes must have either project or work package proposed changes'
+      );
+    } else if (projectProposedChanges) {
```
RChandler234, line 629: you're overcomplicating this, from this point, have one else if for projectProposedChanges and do everything you need for that inside of there (validating any inputs that correspond to models in the database, and then creating the wbsProposedChanges and projectProposedChanges in the database, I think you should be able to do it in one prisma call) then have another else if that does the same for workPackageProposedChanges

src/backend/src/services/change-requests.services.ts
```diff
@@ -614,87 +613,93 @@ export default class ChangeRequestsService {
...
+      ({ name, status, projectLeadId, projectManagerId, links } = projectProposedChanges);
```
RChandler234, line 630: this syntax is wrong, try `const {name, your other stuff} = projectProposedChanges; ` then you shouldn't need to declare the variables before this line (lines 616-620)

RChandler234 (changes requested):

src/backend/src/services/change-requests.services.ts
```diff
@@ -614,87 +612,114 @@ export default class ChangeRequestsService {
...
-            if (!linkType) throw new NotFoundException('Link Type', link.linkTypeName);
+      const createdProposedChanges = await prisma.wbs_Proposed_Changes.create({
+        data: {
+          changeRequestId: createdCR.scopeChangeRequest!.scopeCrId,
+          name,
+          status,
+          projectLeadId,
+          projectManagerId,
+          links: {
+            create: links.map((linkInfo) => ({ url: linkInfo.url, linkTypeName: linkInfo.linkTypeName }))
```
RChandler234, line 658: you should be able to do a similar create for the projectproposedsolutions and do everything for this if statement in one prisma call (same idea for workpackageproposedsolutions)

src/backend/src/services/change-requests.services.ts
```diff
@@ -614,87 +612,114 @@ export default class ChangeRequestsService {
...
+
+      if (projectManagerId) {
+        const projectManager = await prisma.user.findUnique({ where: { userId: projectManagerId } });
+        if (!projectManager) throw new NotFoundException('User', projectManagerId);
+      }
+
+      if (links.length > 0) {
+        for (const link of links) {
+          const linkType = await prisma.linkType.findUnique({ where: { name: link.linkTypeName } });
+          if (!linkType) throw new NotFoundException('Link Type', link.linkTypeName);
```
RChandler234, line 639: since these are all repeated, you can abstract them into a helper method

src/backend/src/utils/validation.utils.ts
```diff
@@ -64,23 +64,19 @@ const wbsProposedChangesExists = (validationObject: ValidationChain): Validation
...
-  intMinZero(body('wbsProposedChanges.projectManagerId').optional())
-];
-
 const projectProposedChangesExists = (validationObject: ValidationChain): ValidationChain => {
   return validationObject.if((value: any, { req }: any) => req.body.projectProposedChanges);
 };
 
 export const projectProposedChangesValidators = [
   body('projectProposedChanges').optional(),
+  nonEmptyString(wbsProposedChangesExists(body('projectProposedChangesExists.name'))),
```
RChandler234, line 73: if you go look at this helper method, it's basically saying, only validate this body item based on these things if this thing is true (in this case, if the body has wbsProposedChanges). Because we're eliminating the wbsProposedChanges field in the route, you should use projectProposedChangesExists in its place for the projectProposedChangesValidators and workPackageProposedChangesExists in its place for the workPackageProposedChangesValidators

src/shared/src/types/change-request-types.ts
```diff
@@ -129,6 +126,11 @@ export interface ProjectProposedChangesCreateArgs {
 }
 
 export interface WorkPackageProposedChangesCreateArgs {
+  name: string;
+  status: WbsElementStatus;
+  projectLeadId: number;
+  projectManagerId: number;
+  links: { url: string; linkTypeName: string }[];
```
RChandler234, line 133: since these are common fields, you can keep them on WBSProposedChangesCreateArgs and just have ProjectProposedChangesCreateArgs and WorkPackageProposedChangesCreateArgs extend WBSProposedChangesCreateArgs

RChandler234 (changes requested):

src/backend/src/utils/validation.utils.ts
```diff
@@ -115,6 +116,16 @@ export const workPackageProposedChangesValidators = [
   nonEmptyString(body('workPackageProposedChanges.deliverables.*'))
 ];
 
+export const validateProposedChangesFields = async (
```
RChandler234, line 119: this is too complicated. It looks like you're trying to abstract checks to see whether an object is undefined, I was suggesting you make a helper method that takes in the ids we're checking and just houses the duplicate code (checking for links, checking for project manager ids, and checking for project lead ids)

RChandler234: This helper method is the type of thing that sounds good in theory, but isn't really good in implementation (as you found out). If you ever run into a situation like this where something just isn't working (especially something like this that's unlike any other methods we have), it's probably because it's not the best solution. You should take a step back, try to think of a simpler approach, or ask someone else to put eyes on your code to make sure the idea is a viable one

RChandler234 (changes requested):

src/backend/src/services/change-requests.services.ts
```diff
@@ -614,87 +612,80 @@ export default class ChangeRequestsService {
...
             }
-          });
-        } else if (workPackageProposedChanges) {
-          const { duration, startDate, stage, expectedActivities, deliverables, blockedBy } = workPackageProposedChanges;
+          }
+        }
+      });
+    } else if (workPackageProposedChanges) {
+      const { name, status, projectLeadId, projectManagerId, links } = workPackageProposedChanges;
+      const { duration, startDate, stage, expectedActivities, deliverables, blockedBy } = workPackageProposedChanges;
```
RChandler234, line 661: same here, should be able to consolidate into one line

RChandler234 (changes requested):

RChandler234 (approved):
LGTM 🔥 good work on this one

PR #2834, opened 2024-09-12
#2764 Create set logo endpoint

### LINKED ISSUE #2764: [Home Page] - Create setLogo Endpoint
### Description

We need to create an endpoint in the organization router to set the logo of an organization in the backend. 

First, we need to update the organization model to include a `logoImageId` as an optional field in the prisma schema. Make sure to run `yarn prisma:migrate` after any changes to the schema to perform a migration.

Then we need to add a post endpoint to the organization router under the route 'organization/logo/update'  that accepts a file upload. Include unit tests and a manual test using postman to make sure everything is working correctly.

### Acceptance Criteria

- [ ] There is a new `logoImageId` field in the organization model
- [ ] There is a new post endpoint in the organization router under the route 'organization/logo/update' 
- [ ] unit tests and manual testing

### Proposed Solution

We already do something very similar in the '/images/update' post endpoint, which you can use as a reference.  For storing images on Finishline we first upload them to google drive, and then store the google file id in the database, so use the uploadFile util function to get the google file id of the image.

### Mocks

_No response_

chpy04:
## Changes

Added a setLogo endpoint to "/organizations/images/update" that lets an admin user update the logo for the organization.

## Notes

Could not manually test uploadFile() function with postman, so manual override was used for this function in screenshots

## Test Cases

Case A: fails if user is not an admin
Case B: updates logo if no previous logo exists
Case C: updates a previously existing logo

## Screenshots

Successful Postman Response (await uploadFile() was replaced with {id: "123"})
logoUploadSuccess

Error Postman Response if no image is supplied
logoUploadError

Closes #2764

### DIFF
--- a/src/backend/src/controllers/organizations.controller.ts
+++ b/src/backend/src/controllers/organizations.controller.ts
@@ -55,11 +55,10 @@ export default class OrganizationsController {
 
   static async setLogoImage(req: Request, res: Response, next: NextFunction) {
     try {
-      const logoFile = req.file || null;
-      if (!logoFile) {
+      if (!req.file) {
         throw new HttpException(400, 'Invalid or undefined image data');
       }
-      const updatedOrg = await OrganizationsService.setLogoImage(logoFile, req.currentUser, req.organization);
+      const updatedOrg = await OrganizationsService.setLogoImage(req.file, req.currentUser, req.organization);
 
       return res.status(200).json(updatedOrg);
     } catch (error: unknown) {
--- a/src/backend/src/controllers/organizations.controller.ts
+++ b/src/backend/src/controllers/organizations.controller.ts
@@ -1,5 +1,6 @@
 import { NextFunction, Request, Response } from 'express';
 import OrganizationsService from '../services/organizations.services';
+import { HttpException } from '../utils/errors.utils';
 
 export default class OrganizationsController {
   static async setUsefulLinks(req: Request, res: Response, next: NextFunction) {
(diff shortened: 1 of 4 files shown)


src/backend/src/controllers/organizations.controller.ts
```diff
@@ -51,4 +52,18 @@ export default class OrganizationsController {
       next(error);
     }
   }
+
+  static async setLogoImage(req: Request, res: Response, next: NextFunction) {
+    try {
+      const logoFile = req.file || null;
```
walker-sean: I don't think you need this line - if `req.file` is not there it will be `undefined `which is also falsy

src/backend/tests/unmocked/organization.test.ts
```diff
@@ -190,4 +190,42 @@ describe('Team Type Tests', () => {
...
+    const file2 = { originalname: 'image2.png' } as Express.Multer.File;
+    it('Fails if user is not an admin', async () => {
+      await expect(
+        OrganizationsService.setLogoImage(file1, await createTestUser(wonderwomanGuest, orgId), organization)
+      ).rejects.toThrow(new AccessDeniedAdminOnlyException('update logo'));
+    });
+
+    it('Succeeds and updates the logo', async () => {
+      const testBatman = await createTestUser(batmanAppAdmin, orgId);
+      (uploadFile as Mock).mockImplementation((file) => {
```
walker-sean, line 205: Since this is mocked please move it to the mocked test file

src/backend/src/services/organizations.services.ts
```diff
@@ -131,4 +131,33 @@ export default class OrganizationsService {
...
+    submitter: User,
+    organization: Organization
+  ): Promise<Organization> {
+    if (!(await userHasPermission(submitter.userId, organization.organizationId, isAdmin))) {
+      throw new AccessDeniedAdminOnlyException('update logo');
+    }
+
+    const logoImageData = await uploadFile(logoImage);
+
+    const newOrg = await prisma.organization.update({
```
walker-sean: Rename to `updatedOrg`

walker-sean (changes requested):
Nice!

src/backend/tests/unmocked/organization.test.ts
chpy04: I don't see a mocked test file in the backend. Should I create one somewhere? Should I also move the other mocked function in this file?

caiodasilva2005 (approved):
LGTM!

PR #2873, opened 2024-10-09
#2803 Change Requests To Review

### LINKED ISSUE #2803: [Home Page] - Unreviewed Change Requests Section
### Description

We want heads/admins to be able to view all their unreviewed change requests for their projects in a central area on the home page to easily check the status of project management. 

Use the ChangeRequestRow component that is already made to make a horizontal list of all the users unreviewed change requests.

The page block should be titled ‘Unreviewed Change Requests’ followed by the number of unreviewed change requests in the list.

### Acceptance Criteria

- [ ] A new UnreviewedChangeRequestDisplay component is added to the lead homepage

### Proposed Solution

We already do something very similar to this on the Change Request Page, so you can use that as reference

### Mocks

image

chpy04:
## Changes

- Added a component that displays any change requests that currently requires a user's review.
- refactored duplicate code for producing change requests into change request utils file

## Notes

Had to override overflow in the`ChangeRequestRow` component to make it scroll. Also, change request cards are somewhat difficult to see but I couldn't change the color while also using the `ChangeRequestRow`, so lmk if I should just display them myself. 

## Screenshots

CRsScroll
CRsEmpty

Closes #2803


src/frontend/src/components/ChangeRequestRow.tsx
```diff
@@ -12,9 +12,10 @@ interface ChangeRequestRowProps {
   title: string;
   changeRequests: ChangeRequest[];
   noChangeRequestsMessage: string;
+  flexWrap?: string;
```
caiodasilva2005: Change this to a boolean and have it be wrap if true and nowrap if false. Just make sure to update the change request row props in the change request overview page with this approach

caiodasilva2005 (changes requested):
Really nice work! Just one small thing

caiodasilva2005 (changes requested):
Changes requested in slack

walker-sean (changes requested):
The  "Unreviewed Change Requests" section on the CR page shows the CRs the user has *made* and is waiting for review on. I think that's what we're looking for here too. If not, we should change the title to match what's on the CR page

chpy04:
> The "Unreviewed Change Requests" section on the CR page shows the CRs the user has _made_ and is waiting for review on. I think that's what we're looking for here too. If not, we should change the title to match what's on the CR page

I was interpreting it as admins will be able to see the change requests that they need to review, but @caiodasilva2005 should confirm that

caiodasilva2005:
@walker-sean yeah I was thinking the same as what Chris mentioned because it would prob be a useful tool for admins to keep track of the change requests they need to review. In this case, yeah I believe the title should be changes to "Change Requests to Review" or something like that. This way, this component will have different functionality than the on the CRs page

PR #2886, opened 2024-10-11
filter out currentUser from CR reviewer options

vsp05:
## Changes
Previously, the current user wasn't being filtered out of CR reviewer options, so they could technically request themselves (at least from a front-end standpoint). The only change was a quick fix to filter out the current user so they can't request themselves on a CR review.


src/frontend/src/pages/ChangeRequestDetailPage/ChangeRequestActionMenu.tsx
```diff
@@ -65,6 +65,10 @@ const ChangeRequestActionMenu: React.FC<ChangeRequestActionMenuProps> = ({
   const isRequestAllowed =
     changeRequest.submitter.userId === currentUser.userId && changeRequest.status === ChangeRequestStatus.Open;
 
+  function potentialCrReviewers(value: UserWithScheduleSettings): value is UserWithScheduleSettings {
```
walker-sean: Can you make this a `const` arrow function so you shouldnt need the `is` type assertion and you dont have to use an arrow function in the filter

walker-sean (changes requested):

walker-sean (approved):

PR #2892, opened 2024-10-22
Add FinishLine RR ID to Finance CSV

vsp05:
Quick feature Alex wanted (adding onto a previous PR I made about adding account codes to the CSV).


Peyton-McKee (changes requested):
This should be the Identifier not the rrid

Peyton-McKee (approved):
This should be the crIdentifier not the rrid

Issue #3044, opened 2024-11-25
[Home Page] - Slack Integration Event Listener


chpy04:
### Description

We need to set up an endpoint that can listen to events from the Slack bot. This endpoint will verify events, process them, and add appropriate notifications to users in the backend. For this ticket, only the reception and verification of events is needed.

### Acceptance Criteria

- An endpoint is set up to listen to events from the slack bot
- the events are verified to confirm that events are from slack
- the events can be processed for future features

### Proposed Solution

Set up an endpoint in express for the events. Add the URL as a callback URL for our slack bot. Upon receiving events, authorize them

PR #3089, opened 2024-12-20
#3044 slack endpoint listener

### LINKED ISSUE #2823: [Home Page] - Receiving Slack Channel Announcements
### Spike Type

Functional

### Goal

We want to look into retrieving messages sent in team slack channels. Specifically, we want to access the messages that ping the channel (@channel), which will serve as our general announcements for the team. We want to retrieve the slack messages from the slack channel that corresponds to the current user's team.
	We have not implemented something like this in Finishline, so this will require some research. Please document your findings with thorough comments and links to resources.

### Reason for Spike

We want our general announcements section on the homepage to be a useful resource.

### Additional notes

_No response_

chpy04:
## Changes

Added slack event listener to listen for user mentions, and add announcements to appropriate users. Announcements are sent when a user is mentioned via a single mention or a broadcast such as @ channel. Event listener also edits and deletes previously sent messages. Users are given announcements based on the slack id in their settings. Additionally, organizations now keep track of the workspace id, so in order to enable the listener an admin must add the slack workspace id to their organization.

## Notes

In the future, we plan to use OAuth to allow users to add their slack to the app so that the bot can listen to multiple organizations. Currently, the bot is only able to listen to NERs slack, and other slack apps would have to be added manually.

## Test Cases

There were multiple test cases for the slack event processing, including but not limited to:
- @ channel, @ here, and @ everyone broadcasts
- parses different kinds of text
- correctly identifies edits and deletes on messages
- ignores unrelated events

Announcement and Organization service functions were also tested

## To Do

- [ ] add endpoint to use setOrganizationSlackWorkspaceId

## Checklist

- [ ] No merge conflicts
- [ ] All checks passing

Closes #2823


src/backend/src/integrations/slack.ts
```diff
@@ -155,4 +155,88 @@ const generateSlackTextBlock = (message: string, link?: string, linkButtonText?:
...
+  try {
+    const channelRes = await slack.conversations.info({ channel: channelId });
+    return channelRes.channel?.name || 'Unknown Channel';
+  } catch (error) {
+    throw new HttpException(500, 'Error getting slack channel name: ' + (error as any).data.error);
+  }
+};
+
+/**
+ * Given a slack user id, prood.uces the name of the channel
```
Peyton-McKee, line 204: fix typos

src/backend/src/routes/slack.routes.ts
```diff
@@ -0,0 +1,19 @@
+import { createEventAdapter } from '@slack/events-api';
+import slackServices from '../services/slack.services';
+import OrganizationsService from '../services/organizations.services';
+import { getWorkspaceId } from '../integrations/slack';
+
+export const slackEvents = createEventAdapter(process.env.SLACK_SIGNING_SECRET || '');
+
+slackEvents.on('message', async (event) => {
+  const organizations = await OrganizationsService.getAllOrganizations();
```
Peyton-McKee: this all needs to be wrapped in a controller, with a try catch, will crash the server otherwise if one of these functions fails

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+  usergroup_id?: string;
+}
+
+export default class slackServices {
+  /**
+   * Converts a SlackRichTextBlock into a string representation for an announcement.
+   * @param block the block of information from slack
+   * @returns the string that will be combined with other block's strings to create the announcement
+   */
+  private static async blockToString(block: SlackRichTextBlock): Promise<string> {
```
Peyton-McKee: think this should be in a util file

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+  }
+
+  /**
+   * Gets the users notified in a specific SlackRichTextBlock.
+   * @param block the block that may contain mentioned user/users
+   * @param usersSettings the settings of all the users in prisma
+   * @param channelId the id of the channel that the block is being sent in
+   * @returns an array of prisma user ids of users to be notified
+   */
+  private static async blockToMentionedUsers(
```
Peyton-McKee: same with this

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+   * make the appropriate announcement change in prisma.
+   * @param event the slack event that will be processed
+   * @param organizationId the id of the organization represented by the slack api
+   * @returns an annoucement if an announcement was processed and created/modified/deleted
+   */
+  static async processMessageSent(event: SlackMessageEvent, organizationId: string): Promise<Announcement | undefined> {
+    let slackChannelName: string;
+    //get the name of the channel from the slack api
+    try {
+      slackChannelName = await getChannelName(event.channel);
```
Peyton-McKee: lets make getChannelName nonthrowing, and it returns either the channel name or undefined, then we just null coalesce this value to use unknow_channel as the default

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+
+    //get the message that will be processed either as the event or within a subtype
+    let eventMessage: SlackMessage;
+
+    if (event.subtype) {
+      switch (event.subtype) {
+        case 'message_deleted':
+          //delete the message using the client_msg_id
+          eventMessage = (event as SlackDeletedMessage).previous_message;
+          try {
```
Peyton-McKee: i think this try catch is unnecessary

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+    const userSettings = await Promise.all(
+      users.map((user) => {
+        return UsersService.getUserSettings(user.userId);
+      })
+    );
+
+    //get the name of the user that sent the message from slack
+    let userName: string = '';
+    try {
+      userName = (await getUserName(eventMessage.user)) ?? '';
```
Peyton-McKee: yeah lets make all these util functions non throwing, and instead return undefined if unable to complete the request

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+    } catch (ignored) {}
+
+    //if slack could not produce the name of the user, look for their name in prisma
+    if (!userName) {
+      const userIdList = userSettings
```
Peyton-McKee: you can replace this with a user.findUnique({ where: { userSettings: { slackId: eventMessage.user } } }) I think.

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+      } else {
+        userName = 'Unknown_User:' + eventMessage.user;
+      }
+    }
+
+    //pull out the blocks of data from the metadata within the message event
+    const richTextBlocks = eventMessage.blocks?.filter((eventBlock: any) => eventBlock.type === 'rich_text');
+
+    if (richTextBlocks && richTextBlocks.length === 1) {
+      for (const element of richTextBlocks[0].elements[0].elements) {
```
Peyton-McKee, line 124: i think we need to check that richTextBlocks[0].elements has a length > 0 as well.

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+          : 'Unknown User:' + eventMessage.user;
+      } else {
+        userName = 'Unknown_User:' + eventMessage.user;
+      }
+    }
+
+    //pull out the blocks of data from the metadata within the message event
+    const richTextBlocks = eventMessage.blocks?.filter((eventBlock: any) => eventBlock.type === 'rich_text');
+
+    if (richTextBlocks && richTextBlocks.length === 1) {
```
Peyton-McKee: why does this have to be exactly one?

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+      for (const element of richTextBlocks[0].elements[0].elements) {
+        messageText += await slackServices.blockToString(element);
+        userIdsToNotify = userIdsToNotify.concat(
+          await slackServices.blockToMentionedUsers(element, userSettings, event.channel)
+        );
+      }
+    } else {
+      return;
```
Peyton-McKee, line 129: add comment saying were disregarding messages without any mentions

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+    //if no users are notified, disregard the message
+    if (userIdsToNotify.length === 0) {
+      return;
+    }
+
+    console.log('processed event');
+
+    if (event.subtype === 'message_changed') {
+      //try to edit the announcement, if no announcement with that id exists create a new announcement
+      try {
```
Peyton-McKee, line 142: get rid of unnecessary try catch

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,289 @@
...
+        messageText,
+        userIdsToNotify,
+        dateCreated,
+        userName,
+        eventMessage.client_msg_id,
+        slackChannelName,
+        organizationId
+      );
+    } catch (error) {
+      //if announcement does not have unique cient_msg_id disregard it
```
Peyton-McKee: this doesnt have to be the case why this statement fails, but we should get rid of this try catch anyways

src/backend/src/services/users.services.ts
```diff
@@ -576,13 +578,34 @@ export default class UsersService {
    * @returns the unread notifications of the user
    */
   static async getUserUnreadNotifications(userId: string, organization: Organization) {
+    const unreadNotifications = await prisma.notification.findMany({
+      where: {
+        users: {
```
Peyton-McKee: i would assume youd also only want notifcations for this organization

Peyton-McKee: and non deleted notifications? or are we hard deleting those

src/backend/src/services/users.services.ts
```diff
@@ -576,13 +578,34 @@ export default class UsersService {
...
+          some: { userId }
+        }
+      },
+      ...getNotificationQueryArgs(organization.organizationId)
+    });
+
+    if (!unreadNotifications) throw new HttpException(404, 'User Unread Notifications Not Found');
```
Peyton-McKee: i think this will throw if this is empty, i dont think we want to do that right, we just want to return the empty list?

src/backend/src/utils/auth.utils.ts
```diff
@@ -31,7 +31,8 @@ export const requireJwtProd = (req: Request, res: Response, next: NextFunction)
   if (
     req.path === '/users/auth/login' || // logins dont have cookies yet
     req.path === '/' || // base route is available so aws can listen and check the health
-    req.method === 'OPTIONS' // this is a pre-flight request and those don't send cookies
+    req.method === 'OPTIONS' || // this is a pre-flight request and those don't send cookies
+    req.path === '/slack' // slack http endpoint is only used from slack api
```
Peyton-McKee, line 35: we should have some sort of authentication for slack requests as well, like how we hvae the notification endpoint auth

Peyton-McKee: are we able to configure a secret to send along with our slackbots requests?

src/backend/tests/mocked/slackMessages.test.ts
```diff
@@ -0,0 +1,464 @@
...
+vi.mock('../../src/integrations/slack', async (importOriginal) => {
+  return {
+    ...(await importOriginal<typeof import('../../src/integrations/slack')>()),
+    getUserName: vi.fn(),
+    getChannelName: vi.fn(),
+    getUsersInChannel: vi.fn()
+  };
+});
+
+describe('Slack message tests', () => {
```
Peyton-McKee: if im being honest, these are still like 90% unmocked tests. If anything we should rename the files to be unit and integration. and this would be integration.

Peyton-McKee (changes requested):
Very very impressive work. Mostly clarifying / cleaning up comments

src/backend/src/routes/slack.routes.ts
chpy04: In theory the package will catch that and throw it to the slackEvents.on('error'... callback, but probably better to wrap it in our controller anyways so will do

src/backend/src/services/slack.services.ts
chpy04: The events can have multiple blocks in them but they would only ever have one with type 'rich_text' so this test is just making sure that one exists. We could change it to > 0 but in theory it would never matter

src/backend/src/services/slack.services.ts
chpy04: This else would indicate that the event we received does not line up with the documentation of the slack API and should never occur, even if no one is mentioned the event will still have those attributes

src/backend/src/services/slack.services.ts
chpy04: My idea behind this try catch was that if we get a message that was edited, but we can't find it in our database, we should create a new message with its content. I think this is an important case as if someone were to make a message but forget to add @ channel, and then edit it to include the broadcast, without this try catch it would get ignored the first time and wouldn't find it so also ignore the edit, but with the try catch it will create a new announcement. LMK if that needs more explaining

src/backend/src/utils/auth.utils.ts
chpy04: the package we use to handle the endpoint does all the authentication automatically

src/backend/src/services/users.services.ts
chpy04: @caiodasilva2005 can you take a look at this?

src/backend/src/services/users.services.ts
chpy04: @caiodasilva2005 this as well

src/backend/src/controllers/slack.controllers.ts
```diff
@@ -0,0 +1,16 @@
+import { getWorkspaceId } from '../integrations/slack';
+import OrganizationsService from '../services/organizations.services';
+import slackServices from '../services/slack.services';
+
+export default class SlackController {
+  static async processMessageEvent(event: any) {
+    try {
+      const organizations = await OrganizationsService.getAllOrganizations();
+      const nerSlackWorkspaceId = await getWorkspaceId();
+      const orgId = organizations.find((org) => org.slackWorkspaceId === nerSlackWorkspaceId)?.organizationId;
```
caiodasilva2005: Small change: I think we should find the organization, and then check if the org exists. The after pass in the organization.organizationId into the function

src/backend/src/services/slack.services.ts
caiodasilva2005: I think what you could do is just make the update announcements service just return if the original announcement is not found, then you just checked if the updated announcement is defined and if it is not then create a new one, then we won't need to use a try catch

src/backend/src/utils/auth.utils.ts
caiodasilva2005: Yeah the api has a built in validation system that invloves sending a challenge parameter to ensure the request is from slack

caiodasilva2005 (changes requested):
After some small changes this is looking really good, great work

src/backend/src/routes/slack.routes.ts
Peyton-McKee: is that true? I’m pretty sure that the event is listening for errors for slackEvents. This is within our API which that slack event i dont believe listens to.

src/backend/src/services/slack.services.ts
Peyton-McKee: so if i did @channel and @here it would only have one rich text?

src/backend/src/services/slack.services.ts
Peyton-McKee: what would the rich text block be then?

src/backend/src/services/slack.services.ts
Peyton-McKee: oh i see. i misread this orgiinally i thought the create announcement function was wrapped inside an else statement. Yeah we should probably be checking for the right error then because it could fail for other reasons that we wouldn’t want to create the announcement in those scenarios

src/backend/src/utils/auth.utils.ts
Peyton-McKee: oh really. so what happens if i try to ping /slack from postman it just fails automatically? thats quite nice

Peyton-McKee: however dont we want tot make sure that its our slackbot thats pinging it. What if somebody elses slack bot tries to reach this endpoint, we shouldn’t allow those connections

src/backend/src/routes/slack.routes.ts
chpy04: that was my understanding of the documentation: https://www.npmjs.com/package/@slack/events-api#handle-errors, but also it can't hurt to have both

src/backend/src/utils/auth.utils.ts
chpy04: We initialize the listener with our signing secret which is then used to authorize events, so if someone else's slack bot tries to reach the endpoint they would have a different signing secret and the authorization would fail

src/backend/src/services/slack.services.ts
chpy04: yes, any message will have one rich_text block, this will then contain within it an array of elements representing the different blocks in the message. Its a bit confusing (and kind of poorly designed if you ask me), but every message has a rich_text block which has a rich_text_section within it which has the elements that we need to parse.

src/backend/src/services/slack.services.ts
chpy04: The slack api documentation is kind of weird in that it doesn't do a great job of describing what attributes show up when, so in all the cases that I tested the rich_text block always showed up in the event, but the API documentation doesn't say that it is a required field, so this was my way of saying if it doesn't have that we will just ignore it.

src/backend/src/utils/auth.utils.ts
Peyton-McKee: oh ok sweet.

src/backend/src/routes/slack.routes.ts
Peyton-McKee: oh wow that is super strange ive never seen an api work like that before.

src/backend/src/services/slack.services.ts
Peyton-McKee: ok sounds good

src/backend/src/services/slack.services.ts
Peyton-McKee: sounds good

src/backend/src/services/users.services.ts
caiodasilva2005: This was addressed in slack so I believe this is all good

src/backend/src/services/users.services.ts
caiodasilva2005: image

It won't throw when it is an empty list, just if its undefined or null. I did a postman test to make sure

src/backend/src/integrations/slack.ts
```diff
@@ -155,4 +155,88 @@ const generateSlackTextBlock = (message: string, link?: string, linkButtonText?:
...
+ */
+export const getChannelName = async (channelId: string) => {
+  const { SLACK_BOT_TOKEN } = process.env;
+  if (!SLACK_BOT_TOKEN) return channelId;
+
+  try {
+    const channelRes = await slack.conversations.info({ channel: channelId });
+    return channelRes.channel?.name || 'Unknown Channel';
+  } catch (error) {
+    return;
```
Peyton-McKee: return undefined.

src/backend/src/integrations/slack.ts
```diff
@@ -155,4 +155,88 @@ const generateSlackTextBlock = (message: string, link?: string, linkButtonText?:
...
+    return members;
+  } catch (error) {
+    return [];
+  }
+};
+
+/**
+ * Given a slack channel id, produces the name of the channel
+ * @param channelId the id of the slack channel
+ * @returns the name of the channel
```
Peyton-McKee: mark the type as returning the name of the channel or undefined if it cannot.

src/backend/src/integrations/slack.ts
```diff
@@ -155,4 +155,88 @@ const generateSlackTextBlock = (message: string, link?: string, linkButtonText?:
...
+ */
+export const getChannelName = async (channelId: string) => {
+  const { SLACK_BOT_TOKEN } = process.env;
```
Peyton-McKee: were not using this token here

src/backend/src/integrations/slack.ts
```diff
@@ -155,4 +155,88 @@ const generateSlackTextBlock = (message: string, link?: string, linkButtonText?:
...
+        members = members.concat(response.members);
+        cursor = response.response_metadata?.next_cursor;
+      } else {
+        throw new Error(`Failed to fetch members: ${response.error}`);
+      }
+    } while (cursor);
+
+    return members;
+  } catch (error) {
+    return [];
```
Peyton-McKee: lets return the members that we may have found already

src/backend/src/integrations/slack.ts
```diff
@@ -155,4 +155,88 @@ const generateSlackTextBlock = (message: string, link?: string, linkButtonText?:
...
+  }
+};
+
+/**
+ * Given a slack user id, prood.uces the name of the channel
+ * @param userId the id of the slack user
+ * @returns the name of the user (real name if no display name)
+ */
+export const getUserName = async (userId: string) => {
+  const { SLACK_BOT_TOKEN } = process.env;
```
Peyton-McKee: same comments here

src/backend/src/integrations/slack.ts
```diff
@@ -155,4 +155,88 @@ const generateSlackTextBlock = (message: string, link?: string, linkButtonText?:
...
+    return;
+  }
+};
+
+/**
+ * Get the workspace id of the workspace this slack api is registered with
+ * @returns the id of the workspace
+ */
+export const getWorkspaceId = async () => {
+  const { SLACK_BOT_TOKEN } = process.env;
```
Peyton-McKee: same comments here

src/backend/src/routes/slack.routes.ts
```diff
@@ -0,0 +1,12 @@
+import { createEventAdapter } from '@slack/events-api';
+import SlackController from '../controllers/slack.controllers';
+
+export const slackEvents = createEventAdapter(process.env.SLACK_SIGNING_SECRET || '');
+
+slackEvents.on('message', async (event) => {
```
Peyton-McKee: you can remove the async (event) and just have it be SlackController.processsMessageEvent

src/backend/src/routes/slack.routes.ts
```diff
@@ -0,0 +1,12 @@
...
+  SlackController.processMessageEvent(event);
+});
+
+slackEvents.on('error', (error) => {
+  console.log(error.name);
```
Peyton-McKee: can we also log the stringified error for more info

src/backend/src/services/announcement.service.ts
```diff
@@ -45,6 +45,66 @@ export default class AnnouncementService {
...
+  }
+
+  static async deleteAnnouncement(slackEventId: string, organizationId: string): Promise<Announcement> {
+    const originalAnnouncement = await prisma.announcement.findUnique({
+      where: {
+        slackEventId
+      }
+    });
+
+    if (!originalAnnouncement) throw new NotFoundException('Announcement', slackEventId);
```
Peyton-McKee, line 96: check if its already deleted

Peyton-McKee: check if its not in this organization

src/backend/src/services/announcement.service.ts
```diff
@@ -45,6 +45,66 @@ export default class AnnouncementService {
...
+    slackChannelName: string,
+    organizationId: string
+  ): Promise<Announcement> {
+    const originalAnnouncement = await prisma.announcement.findUnique({
+      where: {
+        slackEventId
+      }
+    });
+
+    if (!originalAnnouncement) throw new NotFoundException('Announcement', slackEventId);
```
Peyton-McKee, line 63: check if its deleted or not in this organization.

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,179 @@
...
+   * Given a slack event representing a message in a channel,
+   * make the appropriate announcement change in prisma.
+   * @param event the slack event that will be processed
+   * @param organizationId the id of the organization represented by the slack api
+   * @returns an annoucement if an announcement was processed and created/modified/deleted
+   */
+  static async processMessageSent(event: SlackMessageEvent, organizationId: string): Promise<Announcement | undefined> {
+    //get the name of the channel from the slack api
+    const slackChannelName: string = (await getChannelName(event.channel)) ?? `Unknown_Channel:${event.channel}`;
+    const dateCreated = new Date(1000 * Number(event.event_ts));
```
Peyton-McKee, line 80: i thought we were letting the database just make this be the time that it gets created.

src/backend/src/services/slack.services.ts
```diff
@@ -0,0 +1,179 @@
...
+      eventMessage = event as SlackMessage;
+    }
+
+    //loop through the blocks of the meta data while accumulating the
+    //text and users notified
+    let messageText = '';
+    let userIdsToNotify: string[] = [];
+
+    //Get the settings of all users in this organization to compare slack ids
+    const users = await UsersService.getAllUsers();
```
Peyton-McKee: this is getting all of the users in all organizations currently. IMO we should move this entire section into the getUserName function and make a raw prisma query for userSettings which would be more efficient here instead of having to make individual queries for all the users. would also shrink this function a bit which would be nice

Peyton-McKee (changes requested):

src/backend/src/services/slack.services.ts
chpy04: Ok, I can move the logic and prisma queries to work like that. I'm gonna put it into a utils function that gets a user id given a slack id just because then it reduces the amount of work the slack integration function is doing which can't be tested

src/backend/src/routes/slack.routes.ts
chpy04: ok, I will also do that for our controller try-catch to catch errors in our code

src/backend/src/services/announcement.service.ts
chpy04: announcements don't have organizations, so im not sure what this means

Peyton-McKee (approved):

PR #3183, opened 2025-02-02
BOM Drag and Drop To Switch Assemblies

### LINKED ISSUE #3135: [BOM] - drag items in bom to swap assembly
### Description

Be able to drag and drop items in the BOM to swap what assembly it is in, rather than having to use the three dots. You'll need to look into react drag-and-drop functionality (which may have been covered in launchpad).

Yaomister:
## Changes

Implemented so that you can now drag and drop the the materials to change their assemblies. The old button to change assemblies is still kept.
 

## Screenshots

Screenshot 2025-02-02 at 5 56 04 PM
## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3135


src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOMTab.tsx
```diff
@@ -30,13 +30,15 @@ const BOMTab = ({ project }: { project: Project }) => {
...
   const {
     data: materials,
     isLoading: materialsIsLoading,
     isError: materialsIsError,
-    error: materialsError
+    error: materialsError,
+    refetch: refetchMaterials
   } = useGetMaterialsForWbsElement(project.wbsNum);
 
   if (assembliesIsError) return <ErrorPage message={assembliesError.message} />;
```
jonathan-chen10, line 44: can you get rid of the console log on line 49? not your fault but would be good to clear up the console

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/BOMTable.tsx
```diff
@@ -61,10 +64,32 @@ const BOMTable: React.FC<BOMTableProps> = ({ setHideColumn, columns, materials,
     assemblyMaterials.forEach((material, indx) => materialsWithAssemblies.push(materialToRow(material, indx)));
   });
 
+  // drag and drop mechanics
+  const [draggedMaterial, setDraggedMaterial] = useState<Material | null>(null);
```
jonathan-chen10: can you move this to the top of the component where the other useState is?

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/BOMTable.tsx
```diff
@@ -93,13 +118,31 @@ const BOMTable: React.FC<BOMTableProps> = ({ setHideColumn, columns, materials,
         columns={columns as GridColumns<GridValidRowModel>}
         rows={rows.concat(materialsWithAssemblies.filter(isAssemblyOpen))}
         getRowClassName={(params) =>
-          `super-app-theme--${String(params.row.id).includes('assembly') ? 'assembly' : 'material'}`
+          `super-app-theme--${String(params.row.id).includes('assembly') ? `assembly assembly-id:${params.row.assemblyId}` : `material material-id:${params.row.materialId}`}`
```
jonathan-chen10: I don't like injecting the class like this. 

I think you can get the material ID (line 135, 140) using something like `rows.concat(materialsWithAssemblies.filter(isAssemblyOpen))[parseInt(event.currentTarget.dataset.rowindex || '0')].materialId;` where you index the row IDs with the index of the row (but refactor this so it's not one line)

This way you don't need to modify class name

jonathan-chen10 (changes requested):
Good work! Left some feedback for you to address

vsp05:
Great work! If possible, could you make it so that if the user drags the item to the top (the red bar), the item gets removed from all assemblies?

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/BOMTable.tsx
```diff
@@ -61,10 +65,31 @@ const BOMTable: React.FC<BOMTableProps> = ({ setHideColumn, columns, materials,
...
+    assignMaterial(draggedMaterial.materialId, targetAssemblyId)().finally(() => {
+      setDraggedMaterial(null);
+    });
+  };
+
   return (
     <Box
       sx={{
-        height: 'calc(100vh - 200px)',
+        height: 'calc(100vh - 180px)',
```
vsp05: what was the reason for this change?

Yaomister: I don't remember changing this... but I changed it back to 200 regardless

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/BOMTable.tsx
```diff
@@ -100,6 +125,33 @@ const BOMTable: React.FC<BOMTableProps> = ({ setHideColumn, columns, materials,
...
+              }
+              const rowIndex = parseInt(event.currentTarget.getAttribute('data-rowindex') || '0');
+              const materials = rows.concat(materialsWithAssemblies.filter(isAssemblyOpen));
+              const { materialId } = materials[rowIndex];
+              handleDragStart(materialId);
+            },
+            onDrop: (event: React.DragEvent) => {
+              if (event.currentTarget.className.includes('super-app-theme--material')) return;
+              const rowIndex = parseInt(event.currentTarget.getAttribute('data-rowindex') || '0');
+              const materials = rows.concat(materialsWithAssemblies.filter(isAssemblyOpen));
```
jonathan-chen10, line 143: Can you rename specifically this constant? Weird that materials[i] gets you an assembly

jonathan-chen10 (changes requested):
Good work on this ticket! Fix the small issue I had and you should be OK. A few notes:

- you have it so that if the material is dropped anywhere outside the table (not just at the top) it will be unassigned. Double check if that's really the desired functionality - this might make it too easy to accidentally unassign materials.
- sometimes the drag and drop can be a little buggy. Like if a material is unassigned and I drag into an assembly I sometimes need to refresh the page to get it to show in the assembly, or it might not show up in the assembly at all. I had a look and can't figure it out.. should probably be OK to merge first and maybe have a bug ticket later on

jonathan-chen10 (commented):
Thanks for the fix, should be good to go as long as my comments above are addressed

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/BOMTable.tsx
```diff
@@ -9,16 +9,18 @@ import { useState } from 'react';
...
 }
 
-const BOMTable: React.FC<BOMTableProps> = ({ setHideColumn, columns, materials, assemblies }) => {
+const BOMTable: React.FC<BOMTableProps> = ({ setHideColumn, assignMaterial, columns, materials, assemblies }) => {
   const [openRows, setOpenRows] = useState<String[]>([]);
+  const [draggedMaterial, setDraggedMaterial] = useState<Material | null>(null);
 
   const arrowSymbol = (rowId: string) => {
-    return openRows.includes(rowId) ? '▼' : '▶';
+    return openRows.includes(rowId) ? '⮝' : '⮟';
```
vsp05: idk why this changed but can you change it back? This hapened to me in the past once lol

vsp05 (changes requested):
Great work! Yeah, I agree with Jonathan, maybe make it so that it only unassigns when you drag it towards the top

Yaomister:
I made it so that it'll be drop from assemblies when its dragged and dropped to the top. The only issue with the way the BOM table is set up is that the red header (like the part that says status, type, etc) is not considered part of the table... so the drop event doesn't trigger when I drag it there. So how I worked around it was I made it so that if the drop happens between the first assembly and the red header, it picks it up as a drop. I'm not sure if this is the best way to do this, but I think it works. lmk if I need to change this more.

vsp05:
Good work! It's a little buggy I think... When I drag something out of an assembly and then try to drag it back in, it doesn't work sometimes. Do you think you could try to address that? If you're stuck I can take a look.

Yaomister:
Ok let me try some more. I'll let you know tomorrow if I need help

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/BOMTable.tsx
```diff
@@ -138,18 +138,21 @@ const BOMTable: React.FC<BOMTableProps> = ({ setHideColumn, assignMaterial, colu
               handleDragStart(materialId);
             },
             onDrop: (event: React.DragEvent) => {
-              if (event.currentTarget.className.includes('super-app-theme--material')) return;
               const rowIndex = parseInt(event.currentTarget.getAttribute('data-rowindex') || '0');
+              const tableRect = event.currentTarget.getBoundingClientRect();
```
jonathan-chen10: Take a look at this. I think the currentTarget is the row that it lands on (is that always the case?), not the entire table.

jonathan-chen10 (changes requested):
Currently the drag and drop behavior is too buggy to release. Sometimes things go in the wrong places, and sometimes the user will see error Assembly with "assembly-misc" not found. I've highlighted one place you can start looking to see if you at least accurately determine which handleDrop to use.

You can also try refactoring some of this logic into a helper function if it makes sense.

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/BOM/BOMTable.tsx
Yaomister: yeah its always the row, its never the table. Thats the issue I was facing; if I get the currentTarget in onDrop, it gets me the row its dropping on, when I do it in onDragEnd, it gets me the row I'm dragging, so it's never the table.

vsp05 (dismissed):
LGTM! Great work, this proved to be a pretty weird ticket lol. I'll wait for Jonathan to take a look and then we'll request sean's review

yarn.lock
jonathan-chen10, line 1: What's yarn.lock doing here? You shouldn't commit yarn.lock unless you changed a dependency

jonathan-chen10 (changes requested):
Good enough, but remove yarn lock

vsp05 (dismissed):
LGTM! Great work!

jonathan-chen10 (dismissed):
Looks good, thanks for always being on top of this

walker-sean (changes requested):
Found a bug: 
image
After dragging a material - the cost on the old assembly isnt updated

Yaomister:
the price changes when I drag the materials...

walker-sean (changes requested):
Could you make it so that you can drag a material onto the list of materials in an assembly to reassign it as well? Instead of just the header to the assembly

vsp05:
Great work! One small comment - can you make it so that when you drag something to the list of materials at the top, it doesn't display the toast error? I'll attach a screenshot below:
Screenshot 2025-02-26 at 12 04 18 AM

walker-sean (approved):

PR #3208, opened 2025-02-07
3129 bom   if not ready to order then dont require fields

### LINKED ISSUE #3129: [BOM] - If "Not Ready To Order", then don't require fields
### Description
When creating a new entry in a project's Bill of Materials (BOM), it requires the user to fill out almost all of the fields before submitting the modal. However, if the entry is "Not Ready To Order", we often don't know the information that is currently required. Change the current design so that if the status is selected as "Not Ready To Order", the only required field is "Name".

### Screenshots (as needed)

Image

thaninbew:
## Changes

- material type, manufacturer, manufacture part number, quantity, price, subtotal, and linkUrl are now optional fields when NOT_READY_TO_ORDER is selected.
- edit schema to allow optional fields
- edit MaterialForm.tsx to make fields optional when NOT_READY_TO_ORDER is selected
- Changed bom-types.ts to allow optional fields
- edit bom.services.ts to allow optional fields + display N/A or 0, ensures total value is calculated correctly.
- added validation in bom.services.ts to require fields when status isn't NOT_READY_TO_ORDER

## Notes

After writing the tests, I wasn't able to re-run and manually test because I was getting that issue with "no users" again on login and no matter how many times I tried troubleshooting it didn't go away. However, no main logic was changed, I only added validation to help with testing. I also manually tested before I wrote the tests.

## Test Cases
**createMaterial Tests:**
Minimal Fields (NOT_READY_TO_ORDER): Verifies that a material created with NOT_READY_TO_ORDER status sets defaults for optional fields.
All Fields (ORDERED): Checks that creating an ORDERED material with complete details (material type, manufacturer, etc.) stores all values correctly.
Missing Required Fields: Ensures that attempting to create an ORDERED material without mandatory fields throws an appropriate error.

**editMaterial Tests:**

Clearing Fields on Status Change: Confirms that switching a material to NOT_READY_TO_ORDER clears optional fields and applies defaults.
Maintaining Provided Values: Validates that editing a NOT_READY_TO_ORDER material retains updated values when provided.
Status Conversion Requirements: Tests that changing a material from NOT_READY_TO_ORDER to ORDERED fails if required fields are omitted.

## Checklist

- [x ] No linting errors / newline at end of file warnings
- [x ] All code follows repository-configured prettier formatting
- [ x] No merge conflicts
- [x ] All checks passing
- [ x] Remove any non-applicable sections of this template
- [ x] Assign the PR to yourself
- [x ] No `yarn.lock` changes (unless dependencies have changed)
- [x ] Request reviewers & ping on Slack
- [ x] PR is linked to the ticket (fill in the closes line below)

Closes # (issue #)
3129


src/backend/src/routes/projects.routes.ts
```diff
@@ -92,15 +92,50 @@ projectRouter.post(
...
-  nonEmptyString(body('manufacturerPartNumber')),
+  body('materialTypeName').custom((value, { req }) => {
+    if (req.body.status !== 'NOT_READY_TO_ORDER' && !value) {
+      throw new Error('Material Type is required unless status is NOT_READY_TO_ORDER');
+    }
+    return true;
+  }),
+  body('manufacturerName').custom((value, { req }) => {
+    if (req.body.status !== 'NOT_READY_TO_ORDER' && !value) {
+      throw new Error('Manufacturer is required unless status is NOT_READY_TO_ORDER');
```
Peyton-McKee: these are all very similar seems like we would be able to abstract this into a function that just takes in a name for the error message. Also might be better to use Material_Status.NOT_READY_TO_ORDER or whatever its called yk

Peyton-McKee: we need to also maintain the original verification if it exists, looks like we dropped that here.

src/backend/src/services/boms.services.ts
```diff
@@ -75,6 +75,27 @@ export default class BillOfMaterialsService {
     pdmFileName?: string,
     unitName?: string
   ): Promise<Material> {
+    if (status !== Material_Status.NOT_READY_TO_ORDER) {
```
Peyton-McKee: abstract this with edit to a verification function

src/backend/src/services/boms.services.ts
```diff
@@ -110,25 +138,35 @@ export default class BillOfMaterialsService {
 
     if (!perms) throw new AccessDeniedException('create materials');
 
+    const data: any = {
```
Peyton-McKee: this can be of type Prisma.MaterialCreateArgs I believe

src/backend/src/services/boms.services.ts
```diff
@@ -110,25 +138,35 @@ export default class BillOfMaterialsService {
...
+      userCreatedId: creator.userId,
+      name,
+      assemblyId,
+      status,
+      materialTypeId: materialType?.id ?? null,
+      manufacturerId: manufacturer?.id ?? null,
+      pdmFileName: pdmFileName || 'N/A',
+      unitId: unit ? unit.id : null,
+      notes,
+      dateCreated: new Date(),
```
Peyton-McKee: dont need to add this here if you make it default from the schema.

src/backend/src/prisma/schema.prisma
```diff
@@ -720,18 +720,18 @@ model Material {
   userCreated            User            @relation(fields: [userCreatedId], references: [userId], name: "materialCreator")
```
Peyton-McKee, line 720: oh you can add @default(now()) to this dateCreated that should be here

src/backend/src/services/boms.services.ts
```diff
@@ -110,25 +138,35 @@ export default class BillOfMaterialsService {
...
+      wbsElementId: project.wbsElementId,
+      manufacturerPartNumber: manufacturerPartNumber || 'N/A',
+      quantity: quantity || 0,
```
Peyton-McKee: i dont think we want to assign a default value for these, if we are then it wouldnt make sense to change the schema to accept optional values instead, since weve already done the verification above we can just say quantity ?? null and the same for all the other ones. and then remove the if statement below.

src/backend/src/services/boms.services.ts
```diff
@@ -555,7 +593,69 @@ export default class BillOfMaterialsService {
     assemblyId?: string,
     pdmFileName?: string
   ): Promise<Material> {
-    const material = await BillOfMaterialsService.getSingleMaterialWithQueryArgs(materialId, organization);
+    if (status !== Material_Status.NOT_READY_TO_ORDER) {
```
Peyton-McKee: abstract this to a verification function and use it in both create / edit.

src/backend/src/routes/projects.routes.ts
Peyton-McKee: in fact, i think we should remove this since we have the verification inside the service function which i prefer, and rather just add a .isOptional() to the end of the statement like  nonEmptyString(body(‘materialTypeName’).isOptional())

Peyton-McKee: really this route verifications is mainly to make sure the types are right, this is business logic which i believe should be in the service function you know

src/backend/src/services/boms.services.ts
```diff
@@ -555,7 +593,69 @@ export default class BillOfMaterialsService {
...
+        assembly: true,
+        materialType: true,
+        manufacturer: true,
+        wbsElement: true,
+        unit: true
+      }
+    });
+
+    if (!material) throw new NotFoundException('Material', materialId);
+    if (material.dateDeleted) throw new DeletedException('Material', materialId);
```
Peyton-McKee, line 612: can we add an organization check as well

src/backend/src/services/boms.services.ts
```diff
@@ -573,36 +673,34 @@ export default class BillOfMaterialsService {
...
 
-    const materialType = await BillOfMaterialsService.getSingleMaterialTypeWithQueryArgs(materialTypeName, organization);
-
-    let unit = null;
-    if (unitName) {
-      unit = await prisma.unit.findUnique({
-        where: { uniqueUnit: { name: unitName, organizationId: project.wbsElement.organizationId } }
-      });
-      if (!unit) throw new NotFoundException('Unit', unitName);
+    const data: any = {
```
Peyton-McKee: i think you can change this to Prisma.MaterialUpdateArgs it might be called?

src/backend/src/services/boms.services.ts
```diff
@@ -555,7 +593,69 @@ export default class BillOfMaterialsService {
...
+    if (material.dateDeleted) throw new DeletedException('Material', materialId);
+
+    let materialType = null;
+    let manufacturer = null;
+
+    if (materialTypeName) {
+      materialType = await BillOfMaterialsService.getSingleMaterialTypeWithQueryArgs(materialTypeName, organization).catch(
+        () => null
+      );
+      if (status !== Material_Status.NOT_READY_TO_ORDER && !materialType) {
```
Peyton-McKee: we dont want this check i dont think

src/backend/src/services/boms.services.ts
```diff
@@ -555,7 +593,69 @@ export default class BillOfMaterialsService {
...
+        materialType = await BillOfMaterialsService.getSingleMaterialTypeWithQueryArgs(materialTypeName, organization);
+      }
+    }
+
+    if (manufacturerName) {
+      manufacturer = await BillOfMaterialsService.getSingleManufacturerWithQueryArgs(manufacturerName, organization).catch(
+        () => null
+      );
+      if (status !== Material_Status.NOT_READY_TO_ORDER && !manufacturer) {
```
Peyton-McKee: also dont think we need this

src/backend/src/services/boms.services.ts
Peyton-McKee: the reasoning being regardless of whether its not ready to order or not, we want to throw if the id were given isnt valid.

src/backend/src/services/boms.services.ts
Peyton-McKee: i also think this section may be able to be abstracted with create as well

Peyton-McKee (changes requested):
Looks great! nice work!

vsp05:
note: will work on this more after break

PR #3248, opened 2025-02-21
added rrid to slack msg

### LINKED ISSUE #3247: include rrid in slack reimbursement request created
I'm just going to get this ticket out of the way bc I think its too straightforward for any of the devs to waste their time on it

via alex:
Can this line say the ID of the request in it. I know I can open it up, but if I already have a reimbursement open and need to leave a comment. I have to click through to see which opens to the same one. This is applicable as people typically submit all their reimbursements together so I have a bunch that all say the same thing.

Image

vsp05:
Closes #3247

### DIFF
--- a/src/backend/src/services/reimbursement-requests.services.ts
+++ b/src/backend/src/services/reimbursement-requests.services.ts
@@ -196,6 +196,7 @@ export default class ReimbursementRequestService {
 
     await sendReimbursementRequestCreatedNotificationAndCreateMessageInfo(
       createdReimbursementRequest.reimbursementRequestId,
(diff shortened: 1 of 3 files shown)


jonathan-chen10 (changes requested):
check with end user, you probably need to pass ident to the function instead of using the backend uuid

jonathan-chen10 (changes requested):
Pretty good. Can you change the wording so it seems clearer what the number is? Like "ID#{id}" or something along those lines

jonathan-chen10 (approved):
looks good!

PR #3483, opened 2025-06-03
#1192 work package start date bug

### LINKED ISSUE #1192: Backend - Monday Start Date Changes to Tuesday after a certain time of the day in EST
### Description

Theres a very weird bug that after a certain hour of the day, (earliest known occorance is around 7-8pm EST) our backend thinks that the day passed in is actually the next day. 

### Acceptance Criteria

Does not change the start date to a tuesday ever when user selects a monday start date at anytime

### Proposed Solution

Check out the code in create work package file in the backend and look where we manipulate the start date that we pass in. Console log a bunch and see if we can somehow make it so that the date is actually the correct date

### Mocks

_No response_

staysgt:
## Changes

Fixed start date conversion to use a user's local timezone instead of UTC, which was causing the Monday to change to a Tuesday.

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #1192


chpy04 (changes requested):
The JS date object doesn't actually deal with time zones at all, it just stores the number of milliseconds since 1970-01-01 00:00:00 UTC and uses the current time zone to display appropriately. This code just truncates the date object depending on the users time zone, which is bad because then the time zone would have an effect on our backend data, which we want to avoid. Ideally, we store dates as 00:00 UTC of each day, in which case you just need to look at how we are displaying on the frontend, and adjust for the users timezone there instead of modifying the backend data, although you should check that we are actually doing this and there isn't some convertion/truncation similar to this that is causing the problem

PR #3496, opened 2025-07-02
#1784: removed refetching

### LINKED ISSUE #1784: Backend Cleanup - hasBulletCheckingPermissions
### Description

Currently, hasBulletCheckingPermissions is called in the checkDescriptionBullet service method. Currently, we pull a userId off a user then use it to refetch the user inside hasBulletCheckingPermissions. This is a redundant call, so we should just pass in the user instead of refetching it.

### Acceptance Criteria

- [ ] The checkDescriptionBullet in the backend has been updated so that there's not a redundant call to fetch a user inside hasBulletCheckingPermissions
- [ ] Additionally there is some boolean logic that can be simplified in hasBulletCheckingPermissions

### Proposed Solution

This is in the backend, if you have any questions, ask your tech lead!

### Mocks

N/A

hayliepedersen:
## Changes

Passed in user instead of userId to hasBulletCheckingPermissions.

Closes #1784

### DIFF
--- a/src/backend/src/services/description-bullets.services.ts
+++ b/src/backend/src/services/description-bullets.services.ts
@@ -40,7 +40,7 @@ export default class DescriptionBulletsService {
     if (wbsElement.status !== WBS_Element_Status.ACTIVE)
       throw new HttpException(400, 'Cannot check a description bullet on an inactive work package!');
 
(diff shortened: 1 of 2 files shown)


chpy04 (approved):
LGTM

PR #3500, opened 2025-07-14
car number dropdown

walker-sean:
image
Added a simple dropdown for car selection on project form. Also made a quick fix where created wps on create project could have incorrectly typed durations


chpy04 (approved):
lgtm

PR #3507, opened 2025-07-23
Parts Review Fullscreen Option and Hotkeys

chpy04:
## Changes

There is a button to expand the part viewer to be full screen instead of only in a window on the left. Additionally, in review mode, the review buttons are over the top of the window instead of to the left to see more of the part and be more consistent with full screen mode. Finally, hot keys were added so users can control the part review with:
- f: toggle in / out of fullscreen
- cmd or ctrl +/-: zoom in / out of part review
- c: enter / exit comment mode
- m: enter / exit common mistake mode

## Screenshots

Screenshot 2025-07-22 at 11 41 56 PM
Screenshot 2025-07-22 at 11 41 41 PM
Screenshot 2025-07-22 at 11 41 25 PM

### DIFF
--- a/src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
+++ b/src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState } from 'react';
+import React, { useCallback, useEffect, useState } from 'react';
 import { Document, Page, pdfjs } from 'react-pdf';
 import { Box, Typography, IconButton, Grid, Button, Tooltip } from '@mui/material';
 import ZoomInIcon from '@mui/icons-material/ZoomIn';
@@ -24,6 +24,9 @@ import DownloadButton from '../../../components/DownloadButton';
 import { NERButton } from '../../../components/NERButton';
 import ArrowBackIosIcon from '@mui/icons-material/ArrowBackIos';
 import ArrowForwardIosIcon from '@mui/icons-material/ArrowForwardIos';
+import FullscreenIcon from '@mui/icons-material/Fullscreen';
+import FullscreenExitIcon from '@mui/icons-material/FullscreenExit';
+import HelpIcon from '@mui/icons-material/Help';
 import { useCurrentUser } from '../../../hooks/users.hooks';
 import { useToast } from '../../../hooks/toasts.hooks';
 
@@ -82,6 +85,7 @@ const PDFViewer: React.FC<FileDisplayProps> = ({ submission, review, hasNext, ne
   const [position, setPosition] = useState({ x: 0, y: 0 });
   const [isDragging, setIsDragging] = useState(false);
   const [startPos, setStartPos] = useState({ x: 0, y: 0 });
+  const [fullscreen, setFullscreen] = useState(false);
 
   //popups to be displayed (all on submissions, only ones in review on reviews)
   const [variablePopups, setVariablePopups] = useState<Part_Review_Popup[]>([]);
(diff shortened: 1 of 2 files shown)


src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
```diff
@@ -107,6 +111,71 @@ const PDFViewer: React.FC<FileDisplayProps> = ({ submission, review, hasNext, ne
...
+  }, [loadSuccess]);
+
+  const handleZoomOut = useCallback(() => {
+    if (!loadSuccess) return;
+    setScale((prev) => prev / 1.2);
+  }, [loadSuccess]);
+
+  const toggleFullscreen = useCallback(() => {
+    if (!loadSuccess) return;
+    setScale((prev) => prev * (fullscreen ? 0.5 : 2));
```
walker-sean, line 127: Will scale always be doubled when going full screen?

src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
```diff
@@ -107,6 +111,71 @@ const PDFViewer: React.FC<FileDisplayProps> = ({ submission, review, hasNext, ne
```
walker-sean, line 112: Not from this PR but we should get rid of the magic numbers for `editMode`. You can type it using a union type of strings

src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
```diff
@@ -153,7 +222,7 @@ const PDFViewer: React.FC<FileDisplayProps> = ({ submission, review, hasNext, ne
   if (isErrorCommonMistakes) return <ErrorPage error={errorCommonMistakes} />;
 
   //only enter review mode if this user is the creator of the non-complete review
-  const reviewMode = review?.userCreated.userId === user.userId && !review.completedAt;
```
walker-sean, line 156: Why get rid of the const?

walker-sean (changes requested):
Great!

src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
chpy04: Technically it depends on the screen size but that number is just a convenience thing bc on load the pdf viewer makes it fit the non-full-screen box so makes sense to zoom it in a little when they go full screen so users don't have to imediately zoom in every time they full screen

src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
chpy04: bc the useeffect for the key handler needs to know this variable, but we can't determine if its a review or not until after calling the hooks, so instead I initialize it to false at first and then set the value here with it as a dependency in the useeffect

src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
```diff
@@ -347,139 +394,44 @@ const PDFViewer: React.FC<FileDisplayProps> = ({ submission, review, hasNext, ne
...
+          <Box
+            sx={{
+              position: 'absolute',
+              top: 8,
+              right: 8,
+              zIndex: 1
+            }}
+          >
+            <Tooltip
+              title="Press 'f' to toggle full screen, cmd or ctrl +/- to zoom in/out, 'c' to enter common mistake mode, and 'm' to enter common mistake mode"
```
cielbellerose: -> 'c' to enter comment mode

cielbellerose (commented):
looks good to me!

src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
```diff
@@ -107,6 +111,71 @@ const PDFViewer: React.FC<FileDisplayProps> = ({ submission, review, hasNext, ne
...
+        } else if (e.key === '-') {
+          handleZoomOut();
+        }
+      }
+
+      if (e.key === 'f' || e.key === 'F') {
+        toggleFullscreen();
+      }
+
+      if (reviewMode && (e.key === 'c' || e.key === 'C')) {
```
cielbellerose, line 170: typing 'c' or 'm' while writing a comment or custom common mistake moves the marker to the very bottom right. maybe we could add a bool for 'writing in progress' to temp disable the shortcuts?

src/frontend/src/pages/PartPage/PartPageComponents/PdfDisplay.tsx
```diff
@@ -107,6 +111,71 @@ const PDFViewer: React.FC<FileDisplayProps> = ({ submission, review, hasNext, ne
...
   const { mutateAsync: updateReview } = useEditPartReview();
 
+  const handleZoomIn = useCallback(() => {
+    if (!loadSuccess) return;
+    setScale((prev) => prev * 1.2);
+  }, [loadSuccess]);
+
+  const handleZoomOut = useCallback(() => {
+    if (!loadSuccess) return;
+    setScale((prev) => prev / 1.2);
```
cielbellerose, line 122: totally optional but it might be nice to have a maximum for zooming out

```ts
const newScale = prev / 1.2;
return newScale >= 0.1 ? newScale : prev;
```

or possibly a home shortcut to recenter/resize

cielbellerose (changes requested):

cielbellerose (approved):
Looks great!

walker-sean (approved):

PR #3518, opened 2025-08-03
#3516 fixing containerization build

### LINKED ISSUE #3516: [Docker] - E2E tests failing
### Description

Our e2e tests don't pass due to the fact that the backend container starts and then immediately stops when they are run.

### Acceptance Criteria

The e2e test actually passes

### Proposed Solution

Figure out why the prod container is failing immediately. Reference dev containers because those work

### Mocks

_No response_

bpalazzi512:
## Changes

building the backend module wouldn't generate anything because we didn't include the proper files (you can see in the changes). Issues also propped up with the shared module because the module system we were using was interpretting some files as individual modules

## Notes
n/a

## Test Cases

n/a

## Screenshots

n/a

## To Do

verify testing works still

## Checklist

- [ ] No merge conflicts
- [ ] All checks passing

Closes #3516


devContainerization/docker-compose.dev.yml
```diff
@@ -1,7 +1,7 @@
 name: finishline-development
 services:
   database:
-    container_name: database
+    container_name: database-dev
```
chpy04, line 4: Im changing these bc its annoying when you can't run the system tests at the same time bc the names were the same

src/shared/tsconfig.json
```diff
@@ -1,7 +1,7 @@
 {
   "extends": "../../tsconfig.build.json",
   "compilerOptions": {
-    "module": "esnext",
+    "module": "CommonJS",
```
walker-sean, line 4: Why is this changed? We use ES imports in this directory, so I imagine this would cause issues

walker-sean (changes requested):

src/shared/tsconfig.json
chpy04: Short answer is this is what works. The long answer is our backend compiles on CommonJS (which is standard for node apparently), but the frontend is on esnext, so no matter which we compile the shared too it won't match for one of them, and esnext makes the build fail but CommonJS doesn't because apparently the bundler is ok importing into esnext from CommonJS but not the other way around. Im not a huge fan of this, but from what I can tell our solution would be either to conditionally build the shared folder in two different modules for the frontend and backend (which I hate), or to convert the backend to esnext which would require doing some weird things with the import statements in the compiled build (which I also don't like because it would further extend this already massively delayed release). Unless you know of a reason that this wouldn't work I would prefer to trust the tests but lmk what you think

walker-sean: ok thats awesome, this is fine

walker-sean (approved):

PR #3521, opened 2025-08-05
Fix deprecated docker-compose command in start-containers.js

vsp05:
## Changes

See thread [here](https://nu-electric-racing.slack.com/archives/C074FT6AZ9R/p1754364117458549).

Chris (and any other reviewers): could you try deleting your `finishline-development` container and running the `yarn docker:start` command and making sure it compiles correctly on your end? It works for me but if this PR is breaking something this is where it'd break...

Shouldn't have any complications as long as devs are on docker engine 20.10.13 or later (this was released in 2022 so hopefully all our devs are on a more recent version lol)

Fun info from looking into this:
Previously, Docker architecture treated compose as a separate tool, so users would have to download Docker Engine and then a separate tool `Docker Compose`, which included the `docker-compose` binary. On more recent versions of Docker Desktop/Docker Engine, the compose functionality is bundled directly into the Docker CLI as `docker compose`.

### DIFF
--- a/devContainerization/start-containers.js
+++ b/devContainerization/start-containers.js
@@ -7,7 +7,7 @@ const COMPOSE_FILE = 'docker-compose.dev.yml';
 async function getContainerStatus() {
   try {
     //get all containers from the compose file (running and stopped)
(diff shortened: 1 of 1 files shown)


chpy04 (approved):
lgtm

PR #3525, opened 2025-08-08
Get rid of other usages of docker-compose

vsp05:
## Changes
found some other usages of docker-compose in package.json

See #3521 for context


chpy04 (approved):
tbh thought these were just aliases until you said anything lol

PR #3533, opened 2025-08-18
#1066 Deadline Not Being Set Correctly After Creating Task

### LINKED ISSUE #1066: Deadline Not Being Set Correctly on Task Creation
### Browser(s)

Chrome

### Platform(s)

Desktop

### Version(s)

production (main)

### Observed Behavior

When creating a task, deadline is being set to the wrong date (it looks like the day before whatever you set) when you actually create the task.

### Expected Behavior

When creating a task, the created task should have whatever deadline you enter into the deadline field.

### Steps to Reproduce

1. Go to a TaskList
2. Create a Task (make sure to set a deadline)
3. Observe the task display a deadline the day before it's supposed to

### Screenshots (as needed)

_No response_

EddieTheEgg:
## Changes

Updated the way we process UTC dates from backend by removing unnecessary timezone conversion in datePipe. Then for editing deadlines in the future I converted deadline back to user local timestamp so DatePicker can display correctly.

## Notes

datePipe was converting incoming UTC dates using toDateString(), which stripped UTC timezone information. Then when we make a new Date object out of this, we have a local timezone date object. So when converting this back to UTC in the final constructor there's where we have one day off issue. This small change fixed the tasks deadline displays, but not sure if it still works for other features that uses datePipe.

## Screenshots
Screenshot 2025-08-17 at 10 40 57 PM
Screenshot 2025-08-17 at 10 40 53 PM
Screenshot 2025-08-17 at 10 40 49 PM

## Checklist
- [x] No merge conflicts
- [x] All checks passing

Closes #1066

### DIFF
--- a/src/frontend/src/apis/transformers/tasks.transformers.ts
+++ b/src/frontend/src/apis/transformers/tasks.transformers.ts
@@ -9,8 +9,11 @@ import { Task } from 'shared';
 export const taskTransformer = (task: Task): Task => {
   let transformedDeadline: Date | undefined;
   if (task.deadline) {
-    const [year, month, day] = task.deadline.toString().split('T')[0].split('-').map(Number);
-    transformedDeadline = new Date(year, month - 1, day);
+    const date = new Date(task.deadline);
+    const year = date.getUTCFullYear();
+    const month = date.getUTCMonth();
+    const day = date.getUTCDate();
+    transformedDeadline = new Date(year, month, day);
   }
   return {
     ...task,
--- a/src/frontend/src/apis/transformers/tasks.transformers.ts
+++ b/src/frontend/src/apis/transformers/tasks.transformers.ts
@@ -7,10 +7,16 @@ import { Task } from 'shared';
  * @returns Properly transformed task object.
  */
 export const taskTransformer = (task: Task): Task => {
+  console.log('This is the task deadline from backend', task.deadline);
+  let transformedDeadline: Date | undefined;
+  if (task.deadline) {
+    const [year, month, day] = task.deadline.toString().split('T')[0].split('-').map(Number);
+    transformedDeadline = new Date(year, month - 1, day);
+  }
   return {
     ...task,
     dateCreated: new Date(task.dateCreated),
     dateDeleted: task.dateDeleted ? new Date(task.dateDeleted) : undefined,
-    deadline: task.deadline ? new Date(task.deadline) : undefined
+    deadline: task.deadline ? transformedDeadline : undefined
   };
 };
--- a/src/frontend/src/apis/transformers/tasks.transformers.ts
+++ b/src/frontend/src/apis/transformers/tasks.transformers.ts
@@ -7,7 +7,6 @@ import { Task } from 'shared';
  * @returns Properly transformed task object.
  */
 export const taskTransformer = (task: Task): Task => {
-  console.log('This is the task deadline from backend', task.deadline);
   let transformedDeadline: Date | undefined;
   if (task.deadline) {
     const [year, month, day] = task.deadline.toString().split('T')[0].split('-').map(Number);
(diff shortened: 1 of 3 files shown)


src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/TaskList/TaskFormModal.tsx
```diff
@@ -53,7 +53,7 @@ const TaskFormModal: React.FC<TaskFormModalProps> = ({ task, onSubmit, modalShow
       title: task?.title ?? '',
       taskId: task?.taskId ?? '-1',
       notes: task?.notes ?? '',
-      deadline: task?.deadline ?? undefined,
+      deadline: task?.deadline ? new Date(task.deadline.getTime() + task.deadline.getTimezoneOffset() * 60000) : undefined,
```
chpy04: why do you add the getTimezoneOffset here but not in the display for tasks? I assume it would be either both or neither, and I also assume it should be neither because this should already be accounted for in the date pipe when we transform the backend response

src/frontend/src/utils/pipes.ts
```diff
@@ -69,13 +69,13 @@ export const emDashPipe = (str: string) => {
...
- * so to get around that we do the toDateString() of the time and pass it into the Date constructor
- * where the constructor assumes it's in UTC and makes the correct Date object finally
+ * Prisma sends dates in UTC but TypeScript assumes it's in your local time,
+ * so to get around that we convert string dates to Date objects and use timeZone: 'UTC'
+ * in toLocaleDateString where the display assumes UTC and shows the correct date consistently
  */
 export const datePipe = (date?: Date, includeYear = true) => {
   if (!date) return '';
-  date = typeof date == 'string' ? new Date(date) : new Date(date.toDateString());
+  date = typeof date == 'string' ? new Date(date) : date;
```
chpy04: new Date(date.toDateString()) strips the time information from the date so we can only work with a specific day. What is the reasoning behind getting rid of this

chpy04 (changes requested):

src/frontend/src/utils/pipes.ts
EddieTheEgg: I originally removed .toDateString() cause it "fixed" the issue since the timezone helped convert the UTC back to the correct day. However this caused inconsistency like when I used getTimezoneOffset for the TaskFormModal but not the display for tasks. I reverted both changes with a new fix.

EddieTheEgg:
Turns out the issue was the taskTransformer. I ran a deep logging to check from frontend -> backend -> frontend and found the taskTransformer returned one day earlier. 
Screenshot 2025-08-23 at 3 35 17 PM 
Hence why date pipe was getting the wrong Date info as well.

To fix this, I took the deadline from the backend and stripped the time information (assuming we don't care about time info) and just created a new date out of it to return this format:
Correct_Formatting
I didn't want to return the backend deadline format since the original line returned in the Date object format in the pic above, so I took this approach.

src/frontend/src/apis/transformers/tasks.transformers.ts
```diff
@@ -7,10 +7,15 @@ import { Task } from 'shared';
  * @returns Properly transformed task object.
  */
 export const taskTransformer = (task: Task): Task => {
+  let transformedDeadline: Date | undefined;
+  if (task.deadline) {
+    const [year, month, day] = task.deadline.toString().split('T')[0].split('-').map(Number);
```
chpy04: use other date methods to pull out the year month and day instead of parsing the date object

src/frontend/src/utils/pipes.ts
```diff
@@ -75,7 +75,9 @@ export const emDashPipe = (str: string) => {
  */
 export const datePipe = (date?: Date, includeYear = true) => {
   if (!date) return '';
+
```
chpy04: get rid of these changes

chpy04 (changes requested):

PR #3548, opened 2025-08-25
Onboarding Contact Selection is Autocomplete

walker-sean:
The select has become an autocomplete so people dont have to sift through a massive select

image


chpy04 (approved):
lgtm

PR #3557, opened 2025-08-31
#3308: drag to resize graphs

### LINKED ISSUE #3308: [stats] - ability to resize graphs
Be able to resize graphs:

"Ability to resize graphs, some are very small and with like lots of bars it’s very hard to read in such a small area."

Image

vsp05:
## Changes
Implements drag-to-resize functionality for graphs on the stats page. Basically vibe-coded my way through the javascript/react functionality but essentially its just resizing the graphs based on mouse location and movement. Had to adjust the css for a few things just so that the graphs could take on like a more flex display in case of an overflow (which happens when you make graphs bigger/smaller). Also moved the `add graph` button to the top right corner I think it looks better that way and it doesn't get in the way of resizing graphs, but I can move it back if you're particular about it staying there.

Also the graph seed data was a little bit weird so I went ahead and fixed that. 

Lmk if you have any suggestions!

Video:
https://github.com/user-attachments/assets/bcb73bd6-e62c-4004-b300-17ae1750ae6f

Closes #3308

### DIFF
--- a/src/backend/src/prisma/seed.ts
+++ b/src/backend/src/prisma/seed.ts
@@ -793,21 +793,16 @@ const performSeed: () => Promise<void> = async () => {
    * Graphs
    */
 
-  /** Graph 1 */
-  const graph1 = await seedGraph(
-    new Date('12/12/2024'),
-    new Date('12/12/2027'),
-    'new graph',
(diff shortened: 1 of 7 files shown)


chpy04 (approved):
sweet

PR #3609, opened 2025-09-21
#3586 create project rule

### LINKED ISSUE #3586: [RulesDashboard] - Create project rule endpoint
### Description

we need to be able to add a rule to a specific project. The rule itself must already exist, we want to create a Project_Rule with the associated project and rule. This should be under the route `rules/projectRule/create`. 

This will also set the currentStatus to Review (all rules start out as review)

Add seed data for project rule. If createRules #3563 is done there should be rule seed data to use, if not rule seed data may need to be added as well

### Acceptance Criteria

- [ ] createProjectRule route, controller, and service method
- [ ] only admin can create project rules
- [ ] add seed data for project rule
- [ ] unit tests added
- [ ] screenshot of manual tests from postman

### Proposed Solution

createProjectRule route, controller, and service method. for examples to reference, check out the createCar methods or the createGraph methods in statistics

### Mocks

_No response_

cielbellerose:
## Changes

create a project rule so we can add rules to specific projects! controller/service/route, seed data, tests

## Test Cases

- rule must exist!
- rule must not have dateDeleted
- project must exist!
- there must not be a project rule that already exists for this combination
- user must be leadership or above "Project leads/managers and E-Board"

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3586

### DIFF
--- a/src/backend/src/prisma-query-args/rules.query-args.ts
+++ b/src/backend/src/prisma-query-args/rules.query-args.ts
@@ -20,6 +20,7 @@ export const getProjectRuleQueryArgs = () =>
         include: {
           userUpdated: {
             select: {
+              userId: true,
               firstName: true,
               lastName: true
             }
--- a/src/backend/src/prisma-query-args/rules.query-args.ts
+++ b/src/backend/src/prisma-query-args/rules.query-args.ts
@@ -1,1 +1,21 @@
+import { Prisma } from '@prisma/client';
 // write types and query args below here
+
+// preview for rule display
+export const getRulePreviewQueryArgs = () =>
+  Prisma.validator<Prisma.RuleDefaultArgs>()({
+    select: {
+      ruleId: true,
+      ruleCode: true,
+      ruleContent: true,
+      imageFileIds: true
+    }
+  });
+
+export const getProjectRuleQueryArgs = () =>
+  Prisma.validator<Prisma.Project_RuleDefaultArgs>()({
+    include: {
+      rule: getRulePreviewQueryArgs(),
+      project: { select: { projectId: true } }
+    }
+  });
--- a/src/backend/src/prisma-query-args/rules.query-args.ts
+++ b/src/backend/src/prisma-query-args/rules.query-args.ts
@@ -1,5 +1,4 @@
 import { Prisma } from '@prisma/client';
-// write types and query args below here
 
 // preview for rule display
 export const getRulePreviewQueryArgs = () =>
@@ -7,7 +6,7 @@ export const getRulePreviewQueryArgs = () =>
     select: {
       ruleId: true,
       ruleCode: true,
-      ruleContent: true,
+      ruleContent: true
     }
   });
 
@@ -16,7 +15,6 @@ export const getProjectRuleQueryArgs = () =>
     include: {
       rule: getRulePreviewQueryArgs(),
       project: { select: { projectId: true } },
-      currentStatus: true,
       statusHistory: {
         include: {
           userUpdated: {
--- a/src/backend/src/prisma-query-args/rules.query-args.ts
+++ b/src/backend/src/prisma-query-args/rules.query-args.ts
@@ -8,7 +8,6 @@ export const getRulePreviewQueryArgs = () =>
       ruleId: true,
       ruleCode: true,
       ruleContent: true,
-      imageFileIds: true
     }
   });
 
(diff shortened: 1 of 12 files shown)


chpy04:
lol I just wanted to see if copilot would work, im gonna look at this later

src/backend/src/prisma-query-args/rules.query-args.ts
```diff
@@ -1 +1,33 @@
...
+
+export const getProjectRuleQueryArgs = () =>
+  Prisma.validator<Prisma.Project_RuleDefaultArgs>()({
+    include: {
+      rule: getRulePreviewQueryArgs(),
+      project: { select: { projectId: true } },
+      statusHistory: {
+        include: {
+          userUpdated: {
+            select: {
```
chpy04, line 22: good idea to also include userId here as well

src/backend/src/routes/rules.routes.ts
```diff
@@ -7,4 +7,12 @@ const rulesRouter = express.Router();
 
 rulesRouter.post('/rulesetType/create', nonEmptyString(body('name')), validateInputs, RulesController.createRulesetType);
 
+rulesRouter.post(
+  '/projectRule/create',
+  body('ruleId').isString(),
```
chpy04: nonemptystring just for codebase consistency

src/backend/src/services/rules.services.ts
```diff
@@ -25,4 +42,67 @@ export default class RulesService {
...
   }
+
+  static async createProjectRule(
+    submitter: User,
+    organization: Organization,
+    ruleId: string,
+    projectId: string
+  ): Promise<ProjectRule> {
+    if (!(await userHasPermission(submitter.userId, organization.organizationId, isLeadership))) {
+      throw new AccessDeniedException('You do not have permission to create a project rule');
```
chpy04: rename to the action that the user would be trying to accomplish so something like "You do not have permissions to assign rules to projects"

src/backend/src/services/rules.services.ts
```diff
@@ -25,4 +42,67 @@ export default class RulesService {
...
+    }
+
+    const rule = await prisma.rule.findUnique({
+      where: { ruleId },
+      include: {
+        subRules: true,
+        ruleset: { include: { car: { include: { wbsElement: { select: { organizationId: true } } } } } }
```
chpy04: selects instead of includes

src/backend/src/services/rules.services.ts
```diff
@@ -25,4 +42,67 @@ export default class RulesService {
...
+    });
+
+    if (!rule) {
+      throw new NotFoundException('Rule', ruleId);
+    }
+    if (rule.ruleset.car.wbsElement.organizationId !== organization.organizationId) {
+      throw new InvalidOrganizationException('Rule');
+    }
+
+    // must be leaf rule
```
chpy04: Users should be able to assign top level rules to projects. If you want to make it so only leaf rules can be attached to rules then you should recurssively call down to add all the leafs of a top level rule to a project, but I would recommend against that because it is a difficult invarient to hold

chpy04 (changes requested):

src/backend/src/services/rules.services.ts
cielbellerose: in the 'add rule to project' model in the epic it shows only adding leaf rules - all the parents are also added as project rules and show up, but only leaf rules can be selected. now that I think about it we can probably create a separate createProjectLeafRule so when they select the rule all the parents are also added as project rules. but you're right I'll get rid of the leaf stuff for this ticket

chpy04: which model are you refering to? I was looking at the "Users can assign rules by first selecting the team that they want to assign rules to in the left column and then selecting specific rule sections (to select all rules under that rule section) or individual rules." line but I'm not sure if we are talking about different things. I guess my other concern is that you don't actually have a formal difference between leafs and nodes, so theres nothing stopping someone from adding a rule below another one and effectively turning a leaf into a node, which is why Im thinking we don't try to hold this invariant in code and instead just say any rule can be assigned to a project and when you assign a non leaf rule it propogates down and just adds all its children to the project as well

cielbellerose: for project rules “This means that the first field will remain constant by asking them which rule category they want to access (ex. General Regulations), but then it will prompt the user with asking them to select sub-sections under that rule category until they reach a level under which there are individual rules”. I was identifying leafs by if the rule has no subRules. also if users add a nonleaf and it has like 20 child rules does it make sense for them for all of those to be added? esp if its like General Regulations and top level which will have many sections + subsections under it. but idk if its better to keep nonleaf rules empty when theyre added either

chpy04 (approved):

PR #3624, opened 2025-09-26
User endpoints speedup

walker-sean:
A few more endpoints added along the way

I also slimmed down the shared User type


jr1221 (dismissed):
lgtm

src/backend/src/prisma-query-args/tasks.query-args.ts
```diff
@@ -12,3 +13,12 @@ export const getTaskQueryArgs = (organizationId: string) =>
       assignees: getUserQueryArgs(organizationId)
     }
   });
+
+export const getTaskPreviewQueryArgs = () =>
+  Prisma.validator<Prisma.TaskDefaultArgs>()({
+    include: {
+      wbsElement: true,
+      createdBy: true,
+      assignees: true
```
chpy04: probably want to use getUserQueryArgs on these as well

src/backend/src/prisma-query-args/work-packages.query-args.ts
```diff
@@ -49,11 +51,22 @@ export const getWorkPackagePreviewQueryArgs = (organizationId: string) =>
           dateCreated: true,
           dateDeleted: true,
           name: true,
-          lead: getUserQueryArgs(organizationId),
-          manager: getUserQueryArgs(organizationId),
+          lead: { select: { userId: true, firstName: true, lastName: true } },
```
chpy04: I think we should either have a userPreview type that is just the id fname lname or use the getUserQueryArgs. If there is a noticable difference in speed when you do the query args than add the type otherwise I think the select in getUserQueryArgs is achieving basically the same value

src/backend/src/services/users.services.ts
```diff
@@ -42,36 +33,38 @@ import { validateUserIsPartOfFinanceTeamOrHead } from '../utils/reimbursement-re
...
-      return users.map(userWithScheduleSettingsTransformer);
-    }
-
-    const organization = await prisma.organization.findUnique({
-      where: { organizationId },
-      include: {
-        users: getUserWithSettingsQueryArgs(organizationId)
+  static async getAllUsers(): Promise<User[]> {
+    const users = await prisma.user.findMany({
+      select: {
```
chpy04: shouldn't this also use the args?

src/backend/src/services/users.services.ts
```diff
@@ -42,36 +33,38 @@ import { validateUserIsPartOfFinanceTeamOrHead } from '../utils/reimbursement-re
...
+        roles: true,
+        userId: true,
+        firstName: true,
+        lastName: true,
+        email: true
       }
     });
 
-    if (!organization) throw new NotFoundException('Organization', organizationId);
+    users.sort((a, b) => a.firstName.localeCompare(b.firstName));
```
chpy04: use orderby in prisma query

src/backend/src/services/work-packages.services.ts
```diff
@@ -566,4 +569,88 @@ export default class WorkPackagesService {
...
+   * @param organization The organization the current user is logged in for
+   * @param onlyOverdue Whether to only return overdue workpackages
+   */
+  static async getHomePageWorkPackages(
+    user: User,
+    organization: Organization,
+    selection: WorkPackageSelection
+  ): Promise<WorkPackagePreview[]> {
+    const selectionArgs =
+      selection === 'allOverdue'
```
chpy04: use enum values instead of strings

chpy04 (changes requested):

chpy04 (approved):
LGTM

PR #3652, opened 2025-10-13
Concur finance updates

chpy04:
## Changes

all the updates for concur


src/backend/src/routes/organizations.routes.ts
```diff
@@ -3,9 +3,10 @@ import { linkValidators, nonEmptyString, validateInputs } from '../utils/validat
 import OrganizationsController from '../controllers/organizations.controllers';
 import multer, { memoryStorage } from 'multer';
 import { body } from 'express-validator';
+import { MAX_FILE_SIZE } from 'shared';
```
walker-sean, line 6: Do we use the same max file size for everything?

src/backend/src/routes/reimbursement-requests.routes.ts
```diff
@@ -144,8 +152,8 @@ reimbursementRequestsRouter.post(
   nonEmptyString(body('username')).optional(),
   nonEmptyString(body('password')).optional(),
   nonEmptyString(body('discountCode')).optional(),
-  body('taxExempt').isBoolean(),
-  body('twoFactorContacts').isArray(),
+  body('taxExempt').optional().isBoolean(),
+  body('twoFactorContacts').optional().isArray(),
```
walker-sean, line 156: Why make these optional?

src/backend/src/services/organizations.services.ts
```diff
@@ -476,4 +478,71 @@ export default class OrganizationsService {
...
+    const userIdsNoDuplicates = Array.from(new Set(userIds));
+
+    const users = await prisma.user.findMany({
+      where: {
+        userId: { in: userIdsNoDuplicates }
+      }
+    });
+
+    if (users.length !== userIdsNoDuplicates.length) {
+      throw new NotFoundException('User', 'one or more user IDs');
```
walker-sean: This doesn't really fit with the error message it will create, you can just use an httpexception with 404

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -1107,20 +1107,25 @@ export default class ReimbursementRequestService {
...
+      submitter.userId,
+      reimbursementRequest.recipientId
+    );
 
     return reimbursementStatusTransformer(reimbursementStatus);
   }
 
   /**
-   * Adds a reimbursement status with type sabo submitted to the given reimbursement request
+   * Adds a reimbursement status with type pending_sabo_submission to the given reimbursement request
```
walker-sean, line 1138: Update this comment

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -1285,19 +1388,25 @@ export default class ReimbursementRequestService {
...
     if (!existingVendor) {
       throw new NotFoundException('Vendor', vendorId);
     }
 
+    if (existingVendor.organizationId !== organization.organizationId) {
+      throw new InvalidOrganizationException('Vendor');
+    }
+
+    if (existingVendor.addedByUserId !== submitter.userId) {
+      await isUserHeadOrOnFinance(submitter, organization.organizationId);
```
walker-sean: Ik this wasn't you but we should get rid of this function. It does validation rather than being a predicate like everything else it's named like and it's just a wrapper for an actual validation function in the same file

src/backend/src/utils/slack.utils.ts
```diff
@@ -38,6 +38,8 @@ interface SlackMessageThread {
   changeRequestId: string | null;
 }
 
+const DEV_TESTING_OVERRIDE = true;
```
walker-sean: This is a good idea but we could do this as an env variable, either way we shouldn't merge this as true

src/backend/src/utils/slack.utils.ts
```diff
@@ -201,30 +203,45 @@ export const sendThreadResponse = async (threads: SlackMessageThread[], message:
...
 };
 
+export const sendPendingSaboSubmissionNotification = async (
+  threads: SlackMessageThread[],
+  financeUserId: string,
+  pendingSubmissionFromId: string
+) => {
+  await sendThreadResponse(
+    threads,
+    `${await getUserSlackMentionOrName(financeUserId)} has added this reimbursement request to Concur. ${await getUserSlackMentionOrName(pendingSubmissionFromId)}, please approve the request in Concur and mark it as submitted on Finishline.`
```
walker-sean: Maybe we could mention they'll have a link in their email,

src/backend/src/utils/users.utils.ts
```diff
@@ -25,6 +25,20 @@ export const getUserFullName = async (userId: string | null) => {
   return `${user.firstName} ${user.lastName}`;
 };
 
+export const getUserSlackMentionOrName = async (userId: string | null) => {
+  if (!userId) return 'no one';
```
walker-sean: When will userId be null?

src/backend/src/utils/users.utils.ts
```diff
@@ -25,6 +25,20 @@ export const getUserFullName = async (userId: string | null) => {
...
+  const user = await prisma.user.findUnique({
+    where: { userId },
+    include: {
+      userSettings: {
+        select: { slackId: true }
+      }
+    }
+  });
+  if (!user) return 'no one';
```
walker-sean: We probably want to throw here if we do have a userId

src/frontend/src/apis/finance.api.ts
```diff
@@ -196,13 +197,24 @@ export const getAllReimbursements = () => {
...
- * @param id of the reimbursement request being approved by finance
+ * @param id of the reimbursement request being inputted in SABO by finance
+ * @returns the pending sabo submission reimbursement status
+ */
+export const inputReimbursementRequestInSabo = (id: string) => {
+  return axios.post(apiUrls.financeInputReimbursementRequestInSabo(id));
+};
+
+/**
+ * Mark Reimbursement Request as Approved by SABO
```
walker-sean: Wrong comment here

src/frontend/src/hooks/finance.hooks.ts
```diff
@@ -553,17 +556,40 @@ export const useDeleteReimbursementRequest = (id: string) => {
...
+    {
+      onSuccess: () => {
+        queryClient.invalidateQueries(['reimbursement-requests', id]);
+      }
+    }
+  );
+};
+
+/**
+ * Custom react hook to mark a reimbursement request as approved by SABO
```
walker-sean: Here too

src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/ReimbursementRequestDetailsView.tsx
```diff
@@ -353,91 +302,126 @@ const ReimbursementRequestDetailsView: React.FC<ReimbursementRequestDetailsViewP
   const { pathname } = useLocation();
   const urlTabInsert = pathname.includes('/all-requests') ? 'all-requests' : 'my-requests';
 
-  const buttons: ButtonInfo[] = [
+  // Check if the RR has purchase details (receipts and date of expense)
+  const hasPurchaseDetails = reimbursementRequest.receiptPictures.length > 0 && !!reimbursementRequest.dateOfExpense;
+
+  type ButtonInfoWithShow = ButtonInfo & { show?: boolean };
```
walker-sean, line 308: Instead of making a new type, you can add an optional hide value to ButtonInfo, so the same thing can be done in other places easily without changing places that don't need it.

walker-sean: Actually it's not a big deal if we're just filtering here and not conditionally rendering somewhere. So I still prefer it that way but it's not a priority at all

src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/ReimbursementRequestDetailsView.tsx
```diff
@@ -513,6 +498,89 @@ const ReimbursementRequestDetailsView: React.FC<ReimbursementRequestDetailsViewP
...
+      cancelText="No"
+      submitText="Yes"
+      onSubmit={handleMarkPendingFinance}
+    >
+      <CheckList
+        title="Finance Checklist"
+        items={[
+          {
+            resolved: false,
+            detail: `I certify that I have a Concur account${financeDelegates.length > 0 ? ', and have assigned ' + financeDelegates.map(fullNamePipe).join(', ') + ' as delegate(s) to submit this reimbursement request on my behalf.' : '.'}`,
```
walker-sean: Nit: they're not gonna read this anyway but we should say prepare not submit so they know they'll still have to submit

src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/SubmitToSaboModal.tsx
```diff
@@ -53,8 +53,8 @@ const SubmitToSaboModal = ({ open, setOpen, reimbursementRequest }: SubmitToSabo
     setOpen(false);
   };
 
-  const treasurerName = 'Alex Leblang';
-  const treasurerEmail = 'leblang.a@northeastern.edu';
+  const treasurerName = 'Andrew Berkovich';
```
walker-sean, line 57: Do we track treasurer in the DB? The way we've done this will not work in multitenancy

walker-sean (changes requested):

src/backend/src/routes/organizations.routes.ts
chpy04: yes bc at the moment everything is either pdfs or images so there is no reason for it to be different

src/backend/src/routes/reimbursement-requests.routes.ts
chpy04: bc I wanted to make everything that wasn't required so only the field that we have a field for on the frontend is required in the post request although I could easily change these and set defaults so lmk

src/backend/src/services/reimbursement-requests.services.ts
chpy04: I think that is what its doing

src/backend/src/routes/reimbursement-requests.routes.ts
walker-sean: Sounds good

PR #3657, opened 2025-10-14
#3209 sabo number limit

### LINKED ISSUE #3209: [SABO] - leading zeros don't count towards sabo, sabo number doesn't automatically update
### Observed Behavior

When you click "add sabo number" for a Reimbursement Request, there are a couple buggy functionalities.
1. A SABO number is supposed to be 5 digits, but for some reason if you enter in a SABO number with leading zeroes, you're able to type in more than five digits. For example, you can type 011111, and it will just register as 11111.
2. When you submit the "add sabo number" modal, the SABO number box on the reimbursement request details view doesn't update until you refresh the page.

### Expected Behavior

1. only allow 5 digit sabo numbers, even if they start with a 0
2. Automatically update the sabo number on the reimbursement request details view without the user having to refresh the page after submitting the "add sabo number" modal.

jasonk55:
## Changes

Allowed leading zeros to be inputted validly as the SABO number in the SABO number modal.

Added a new pipe that formats the SABO ID to display correctly, padding the number with zeros, up to 5 digits.

## Notes

The pipe I made only pads up to 5 digits as that is the minimum amount of digits needed for a SABO ID.

## Test Cases

- Tested with no leading zeros
- Tested with one leading zero
- Tested with only zeros

## Screenshots

image

image

image

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3209

### DIFF
--- a/src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/AddSABONumberModal.tsx
+++ b/src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/AddSABONumberModal.tsx
@@ -7,12 +7,15 @@ import ReactHookTextField from '../../../components/ReactHookTextField';
 import { useSetSaboNumber } from '../../../hooks/finance.hooks';
 import { useToast } from '../../../hooks/toasts.hooks';
 
-const schema = yup.object().shape({
+const schema = yup.object({
   saboNumber: yup
     .number()
     .typeError('The SABO number should be a valid number')
-    .required()
-    .test('length', 'The SABO number must be at least 5 digits', (num) => String(num).length >= 5)
+    .required('The SABO number is required')
+    .test('exact-5-digits', 'The SABO number must be exactly 5 digits', function () {
+      const original = this.originalValue?.toString().trim();
+      return /^\d{5}$/.test(original || '');
+    })
 });
 
 interface AddSABONumberModalProps {
--- a/src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/AddSABONumberModal.tsx
+++ b/src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/AddSABONumberModal.tsx
@@ -12,9 +12,9 @@ const schema = yup.object({
     .number()
     .typeError('The SABO number should be a valid number')
     .required('The SABO number is required')
-    .test('exact-5-digits', 'The SABO number must be at least 5 digits', function () {
+    .test('exact-5-digits', 'The SABO number must be exactly 5 digits', function () {
       const original = this.originalValue?.toString().trim();
-      return /^\d{5,}$/.test(original || '');
+      return /^\d{5}$/.test(original || '');
     })
 });
 
--- a/src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/AddSABONumberModal.tsx
+++ b/src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/AddSABONumberModal.tsx
@@ -12,9 +12,9 @@ const schema = yup.object({
     .number()
     .typeError('The SABO number should be a valid number')
     .required('The SABO number is required')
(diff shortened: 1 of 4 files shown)


src/frontend/src/pages/FinancePage/ReimbursementRequestDetailPage/AddSABONumberModal.tsx
```diff
@@ -7,12 +7,15 @@ import ReactHookTextField from '../../../components/ReactHookTextField';
...
+const schema = yup.object({
   saboNumber: yup
     .number()
     .typeError('The SABO number should be a valid number')
-    .required()
-    .test('length', 'The SABO number must be at least 5 digits', (num) => String(num).length >= 5)
+    .required('The SABO number is required')
+    .test('exact-5-digits', 'The SABO number must be at least 5 digits', function () {
+      const original = this.originalValue?.toString().trim();
+      return /^\d{5,}$/.test(original || '');
```
gcooper407: I know the original test message indicates otherwise, but a SABO number is supposed to be 5 digits (not just at least 5), so that's what we want to enforce here

src/frontend/src/utils/pipes.ts
```diff
@@ -209,3 +209,14 @@ export const labelPipe = (label: string) => {
 
   return result;
 };
+
+// Pad SABO ID with leading zeroes
+export const formatSaboIdPipe = (saboId: number | undefined): string => {
```
gcooper407, line 214: I think I see what you might be going for here, but it feels like more of a band-aid rather than addressing the underlying issue.

The goal of this ticket is not to pad SABO ids with leading zeroes -- rather, we want to fix the issue where entering a SABO number that starts with zeroes allows users to enter SABO numbers of incorrect lengths. This means that we should be enforcing rules on users' entries before they submit (validating input) rather than modifying their entires after they submit (sanitizing output).

For instance:
`00001234` -- this _should not_ be a valid SABO number, but currently (with your pipe), it gets treated as a valid number (`01234`)

Instead of trying to make some sort of pipe that sanitizes the output, focus on perhaps tweaking the schema object to better validate the input.

gcooper407 (changes requested):
This is a solid start, but I think you might have gotten a bit confused along the way -- but no worries!! It might be helpful to re-read the ticket and play around with the dashboard a bit to visualize the issue, and of course, reach out to me or Harish if you have any questions/need any help!

Your zero-padding pipe doesn't really address the underlying leading zeroes issue outlined in the ticket, and you're also missing a fix for the issue where the "add SABO number" button doesn't update the SABO number box until the page is refreshed. Take some time to try again, and let us know how we can help!

src/frontend/src/utils/pipes.ts
jasonk55: This padding is for visual because if the SABO number is 00001 for example. it would show as just 1 on the RR table. So, I had it pad the zeros to make it visually clear what the number was.

If that's not the way to do it, I was also thinking of taking in the SABO number as a String, but that had some issues with the resolver, so I didn't go down that path. I can look into implementing that, so LMK if that's a better way to handle it.

gcooper407: @jasonk55 I agree that we want `00001` to display as `00001`, but we don't want `00001234` to display as `01234` -- in fact, we don't want it to be allowed at all, but it currently is. So padding the output is a error-prone way of patching this up.

Taking in the SABO number as a String is definitely a step in the right direction (treating them as a number is the whole reason why leading zeroes are causing issues in the first place!) -- play around with that and see if you can get that to work :)

jasonk55: @gcooper407 With the recent commit, would limiting the SABO number to only 5 digits be sufficient, as a user couldn't input `00001234` with this change?

gcooper407: yeah it's fine lol

gcooper407 (approved):
Much better, LGTM

chpy04 (approved):

Issue #3577, opened 2025-09-07
[Calendar Improvements] - editEvent endpoint


wavehassman:
### Description

Implement editEvent endpoint. Edits the Event associated with the given eventId. Pass eventId in the request parameters. The user should be able to edit the:
title,
eventTypeId,
approved?,
approvalByUserId?,
memberIds?,
location?,
zoomLink?,
shopIds,
machineryIds,
workPackageIds,
documentIds,
description?,
daysOfTheWeek, startTime, endTime, recurrenceNumber, initialDateScheduled, allDay (for array of schedule slot),
availability Int[], dateSet (for array of availability)

### Acceptance Criteria

editEvent route, controller, and service method
Added seed data
Added tests

### Proposed Solution

1. Route
2. Controller
3. Service
4. Data
5. Tests

Issue #3570, opened 2025-09-06
[Calendar Improvements] - deleteEvent endpoint


wavehassman:
### Description

Implement deleteEvent endpoint. Soft delete the Event associated with the given eventId. Pass eventId in the request parameters.

### Acceptance Criteria

deleteEvent route, controller, and service method
Added seed data
Added tests

### Proposed Solution

1. Route
2. Controller
3. Service
4. Data
5. Tests

PR #3679, opened 2025-10-27
#3569 final endpoints

### LINKED ISSUE #3569: [Calendar Improvements] - deleteEventType endpoint
### Description

Implement deleteEventType endpoint. Throw exception if a user is not an admin. If so, soft delete the EventType associated with the given eventTypeId. Pass eventTypeId in the request parameters.

### Acceptance Criteria

deleteEventType route, controller, and service method
Added seed data
Added tests

### Proposed Solution

1. Route
2. Controller
3. Service
4. Data
5. Tests

### Mocks

_No response_

wavehassman:
## Changes

Implemented delete and edit event, and delete event type.
I took availabilities out of events because they should be updated automatically on a per user basis. I made the events more like design reviews since we will have to integrate those together. 

## To Do
- make design reviews and meetings event types, and make events to prevent data loss (another ticket)
- slack integration????

## Test Cases

Edit event
- fails if event does not exist
- fails if event is already deleted
- fails if eventTypeId does not exist
- fails if eventType is deleted
- fails if eventType belongs to different organization
- fails if memberIds are invalid
- fails if shopIds are invalid
- fails if shopIds are deleted
- fails if machineryIds are invalid
- fails if machineryIds are deleted
- fails if workPackageIds are invalid
- fails if approvedByUserId is invalid
- succeeds for admin and updates event
- succeeds and updates with minimal fields

Delete event
- fails if event does not exist
- fails if event is already deleted
- succeeds for admin and soft deletes event

Delete EventType
- fails if user is not an admin
- fails if event type does not exist
- fails if event type is already deleted
- fails if event type belongs to different organization
- succeeds for admin and soft deletes event type

## Screenshots
Screenshot 2025-10-26 at 8 29 06 PM
Screenshot 2025-10-26 at 8 28 17 PM
Screenshot 2025-10-26 at 8 29 59 PM

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3569 #3570 #3577


src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +388,273 @@ export default class CalendarService {
...
+   * @param approved Determines if the event has been approved.
+   * @param approvedByUserId The ID of the approving user.
+   * @param questionDocument The link to the question document.
+   * @param location Location of the event.
+   * @param zoomLink Zoom Link if the event is online.
+   * @param description Describes the event.
+   *
+   * @returns The edited event.
+   *
+   * @throws AccessDeniedAdminOnlyException If the submitter is not an admin.
```
JoshuaGoldberg: Is this comment relevant?

JoshuaGoldberg (commented):
Looks great! Only saw a minor comment that might need fixing.

src/backend/src/services/calendar.services.ts
wavehassman: great point

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,272 @@ export default class CalendarService {
...
+          availability: availability.availability,
+          dateSet: availability.dateSet,
+          scheduleSettingsId: scheduleSettings.drScheduleSettingsId
+        };
+      })
+    );
+
+    // Use transaction for the update
+    const updatedEvent = await prisma.$transaction(async (tx) => {
+      await tx.scheduleSlot.deleteMany({
```
chpy04: why do we delete scheduleslots and availabilities any time an event is updated?

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,272 @@ export default class CalendarService {
...
+   */
+  static async deleteEvent(submitter: User, eventId: string, organization: Organization): Promise<Event> {
+    const event = await prisma.event.findUnique({
+      where: { eventId }
+    });
+
+    if (!event) throw new NotFoundException('Event', eventId);
+    if (event.dateDeleted) throw new DeletedException('Event', eventId);
+
+    const hasPermission = await userHasPermission(submitter.userId, organization.organizationId, isAdmin);
```
chpy04: creator of the event should also be able to delete it

chpy04 (changes requested):

src/backend/src/services/calendar.services.ts
wavehassman: I made it so they are only deleted and recreated if there are changes

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,327 @@ export default class CalendarService {
...
+    title: string,
+    eventTypeId: string,
+    organization: Organization,
+    memberIds: string[],
+    shopIds: string[],
+    machineryIds: string[],
+    workPackageIds: string[],
+    documentIds: string[],
+    scheduleSlot: ScheduleSlotCreateArgs[],
+    availability: AvailabilityCreateArgs[],
```
chpy04: we shouldn't be editting the availabilities of people by editting the event. This should be a seperate endpoint

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,327 @@ export default class CalendarService {
...
+    approved: boolean,
```
chpy04: approving an event should be a seperate endpoint that is hit when someone approves the event, this is similar to how the RR endpoints work

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,327 @@ export default class CalendarService {
...
+    approvedByUserId?: string,
```
chpy04: same thing for this

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,327 @@ export default class CalendarService {
...
+   * @returns The edited event.
+   *
+   * @throws NotFoundException If the given event type, member IDs, shop IDs, machinery IDs, work package IDs, document IDs, or approvedByUserId are not found.
+   * @throws InvalidOrganizationException If the given event type, members, shops, machinery, work packages, or approvedByUserId are not part of the same organization.
+   */
+  static async editEvent(
+    submitter: User,
+    eventId: string,
+    title: string,
+    eventTypeId: string,
```
chpy04: is there a reason we support changing the event type? If we want to do this, you would have to verify that the information matches the new event type and get rid of the previous information in the places were the fields differ, but unless there is somewhere in the epic that makes you think we need this I think its fine to just say users have to delete an event and make a new one if they want a different type

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,327 @@ export default class CalendarService {
...
+    organization: Organization,
+    memberIds: string[],
```
chpy04: There should be much looser permissions for adding people to an event than there are for updating the event, so I think this should be a separate endpoint

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,327 @@ export default class CalendarService {
...
+          });
+        }
+
+        return {
+          availability: availability.availability,
+          dateSet: availability.dateSet,
+          scheduleSettingsId: scheduleSettings.drScheduleSettingsId
+        };
+      })
+    );
```
chpy04: you also need to validate that the fields being added / updated actual match the fields in the event type. This should also be done on the create if it isn't already

src/backend/src/services/calendar.services.ts
```diff
@@ -388,6 +387,327 @@ export default class CalendarService {
...
+      // Update the event with new data
+      return await tx.event.update({
+        where: { eventId },
+        data: {
+          title,
+          eventTypeId,
+          members: {
+            set: memberIds.map((userId) => ({ userId }))
+          },
+          shops: {
```
chpy04, line 769: we should ensure at some point that all the machinerys' shops exist in the shops on the event. For example we should never have a machinery on an event but the machineries shop isn't on the event. I think this could either be a validation where you enforce that the shops are also in the shop ids, or you could just add any missing shops to the shopids array here, either works so up to you

chpy04 (changes requested):

src/backend/src/prisma/schema.prisma
```diff
@@ -93,7 +93,7 @@ enum Material_Status {
   READY_TO_ORDER
 }
 
-enum Design_Review_Status {
+enum Event_Status {
```
chpy04, line 96: are we thinking that all events will have a status that is maintained? If so this is fine but if not we could always have designreviewstatus be an optional field on events, and it would only exist for design reviews. Or if you just want to say non DRs go to Scheduled right away thats fine but you would still need some way to determine if an event is a DR bc event types are rows in a table not hard coded

src/backend/src/services/calendar.services.ts
```diff
@@ -249,16 +261,76 @@ export default class CalendarService {
...
+    }
+    if (foundEventType.documents && documentIds.length === 0) {
+      throw new InvalidEventTypeConfigurationException('at least one document');
+    }
+    if (foundEventType.description && !description) {
+      throw new InvalidEventTypeConfigurationException('a description');
+    }
+    if (foundEventType.initialDateScheduled && scheduleSlot.length === 0) {
+      throw new InvalidEventTypeConfigurationException('at least one schedule slot');
+    }
```
chpy04: For all of these, you will need to make a determination of whether or not it is actually rewuired. For example even if an event allows optionalMembers we shouldn't error if the user doesn't add optional members. Also, you need to check the inverse on all of these. If an event type doesn't include something, we should throw an exception if the endpoint tries to supply that thing

src/backend/src/services/calendar.services.ts
```diff
@@ -396,13 +445,9 @@ export default class CalendarService {
...
           }))
         },
-        availabilities: {
-          createMany: {
-            data: availabilitiesWithScheduleSettings
-          }
-        },
-        approved,
-        approvedByUserId,
+        status: Event_Status.UNCONFIRMED,
```
chpy04: not everything should start as unconfirmed because not every event should need to be confirmed. One aditional option on top of what I said above would be to have a "requiresConfirmation" field which determines if this is used. All valid options

src/backend/src/services/calendar.services.ts
```diff
@@ -414,6 +459,529 @@ export default class CalendarService {
...
+    }
+    if (foundEventType.documents && documentIds.length === 0) {
+      throw new InvalidEventTypeConfigurationException('at least one document');
+    }
+    if (foundEventType.description && !description) {
+      throw new InvalidEventTypeConfigurationException('a description');
+    }
+    if (foundEventType.initialDateScheduled && scheduleSlot.length === 0) {
+      throw new InvalidEventTypeConfigurationException('at least one schedule slot');
+    }
```
chpy04: abstract these checks for updates and creates into a helper

src/backend/src/services/calendar.services.ts
```diff
@@ -414,6 +459,529 @@ export default class CalendarService {
...
+    if (foundEventType.name === 'Design Review') {
```
chpy04: Can't do this because we never want to hard code functionality absed on a user input. This is where I actually think the requires confirmation would be a good way to go about this, but you would just need to check the epic vs current DRs and make sure that you don't have other DR specific functionality

src/backend/src/services/calendar.services.ts
```diff
@@ -414,6 +459,529 @@ export default class CalendarService {
...
+   */
+  static async approveEvent(submitter: User, eventId: string, organization: Organization): Promise<Event> {
+    const event = await prisma.event.findUnique({
+      where: { eventId }
+    });
+
+    if (!event) throw new NotFoundException('Event', eventId);
+    if (event.dateDeleted) throw new DeletedException('Event', eventId);
+
+    const hasPermission = await userHasPermission(submitter.userId, organization.organizationId, isAdmin || isHead);
```
chpy04: pretty sure isAdmin || isHead will always evaulate to just isAdmin because functions are truthy

src/backend/src/services/calendar.services.ts
```diff
@@ -414,6 +459,529 @@ export default class CalendarService {
...
+   * @param organization The organization for which the event is being deleted.
+   *
+   * @returns The approved event.
+   *
+   * @throws NotFoundException If the given eventId is not found.
+   * @throws InvalidOrganizationException If the given eventId is not part of the same organization.
+   * @throws DeletedException If the event has already been deleted.
+   * @throws AccessDeniedAdminOnlyException If the submitter is not an admin or head.
+   */
+  static async approveEvent(submitter: User, eventId: string, organization: Organization): Promise<Event> {
```
chpy04, line 813: The businessLogic needs to be baked into the backend not the frontend. For approvals, I think the easiest thing to do is have the approvedby be the owner of the conflicting event when it was made (in which case approved would be false to start). Then the permissions for this would be admin head or the owner of the other meeting. You might want to rename this var then bc right now it looks like it would only exist if someone has approved it.

src/backend/src/services/calendar.services.ts
```diff
@@ -396,13 +445,9 @@ export default class CalendarService {
...
         },
-        availabilities: {
-          createMany: {
-            data: availabilitiesWithScheduleSettings
-          }
-        },
-        approved,
-        approvedByUserId,
+        status: Event_Status.UNCONFIRMED,
+        approved: false,
```
chpy04: approved will require business logic. If there is a conflicting event already in the db (based on whatever the epic says), approved will be false. Otherwise it should just be true. In addition to this if you want to do what I suggested in a comment below and have the approvedByUser be the owner of the other meeting that would also be an option

src/backend/src/services/calendar.services.ts
```diff
@@ -414,6 +459,529 @@ export default class CalendarService {
...
+   * @param location Location of the event.
+   * @param zoomLink Zoom Link if the event is online.
+   * @param description Describes the event.
+   *
+   * @returns The edited event.
+   *
+   * @throws NotFoundException If the given event type, member IDs, shop IDs, machinery IDs, work package IDs, document IDs, or approvedByUserId are not found.
+   * @throws InvalidOrganizationException If the given event type, members, shops, machinery, work packages, or approvedByUserId are not part of the same organization.
+   */
+  static async editEvent(
```
chpy04, line 478: if the date / time of the event is changed, it should:
- go back to being unconfirmed if it is using the DR enums
- check for conflicts with other events, set approved acordingly

chpy04 (changes requested):

src/backend/src/prisma/schema.prisma
wavehassman: Made a requiresConfirmation field, if it requires confirmation it is set to unconfirmed otherwise confirmed. This fixes the hardcoded design review check for the question document, now checks if it requires confirmation.

src/backend/src/services/calendar.services.ts
wavehassman: does not check teams and optional members

src/backend/src/services/calendar.services.ts
wavehassman: what naisha said: Events should only need to be approved if someone else has already booked the same time/place and the person that does the approval should be the person that booked/scheduled a meeting first, so I added that logic

src/backend/src/prisma-query-args/event.query-args.ts
```diff
@@ -11,13 +11,15 @@ export const getEventQueryArgs = (organizationId: string) =>
   Prisma.validator<Prisma.EventDefaultArgs>()({
     include: {
       userCreated: getUserQueryArgs(organizationId),
-      members: getUserQueryArgs(organizationId),
+      requiredMembers: getUserQueryArgs(organizationId),
+      optionalMembers: getUserQueryArgs(organizationId),
+      confirmedMembers: getUserWithSettingsQueryArgs(organizationId),
+      deniedMembers: getUserQueryArgs(organizationId),
       teams: getTeamQueryArgs(organizationId),
```
chpy04: this should only be the information necessary to display the team on the event, so pretty much just name and id

src/backend/src/prisma-query-args/event.query-args.ts
```diff
@@ -11,13 +11,15 @@ export const getEventQueryArgs = (organizationId: string) =>
...
       shops: getShopQueryArgs(organizationId),
       machinery: getMachineryQueryArgs(organizationId),
       workPackages: getWorkPackageQueryArgs(organizationId),
```
chpy04: same here, only the stuff you will actually be using on the frontend

src/backend/src/prisma/schema.prisma
```diff
@@ -1052,18 +1055,23 @@ model Event {
   eventTypeId      String
   eventType        EventType      @relation(fields: [eventTypeId], references: [eventTypeId])
   approved         Boolean        @default(false)
+  // when approval is false, approved by holds who needs to approve the event
+  // when approval is true, approved by holds who actually approved the event
```
chpy04, line 1056: sure but don't call it approvedBy then, maybe something like ApprovalRequiredFrom

src/backend/src/utils/calendar.utils.ts
```diff
@@ -9,3 +12,195 @@ export function buildScheduledTimesOverlap(start?: Date, end?: Date): Prisma.Sch
...
+  if (eventType.requiredMembers && eventData.requiredMemberIds.length === 0) {
+    throw new InvalidEventTypeConfigurationException('at least one required member');
+  }
+  if (eventType.location && !eventData.location) {
+    throw new InvalidEventTypeConfigurationException('a location');
+  }
+  if (eventType.zoomLink && !eventData.zoomLink) {
+    throw new InvalidEventTypeConfigurationException('a zoom link');
+  }
+  if (eventType.shop && eventData.shopIds.length === 0) {
```
chpy04: I don't think that an event type having shop be true should mean it always must have a shop. For most of the things where it is a list, I think it makes more sense for it to be optional to be more flexible for the user, although I will leave this up to you to make the final decision

chpy04 (changes requested):

PR #3736, opened 2025-11-08
Company Table Redesigns

walker-sean:
image
image


src/backend/src/routes/reimbursement-requests.routes.ts
```diff
@@ -167,7 +172,7 @@ reimbursementRequestsRouter.post(
   body('taxExempt').optional().isBoolean(),
   body('twoFactorContacts').optional().isArray(),
   nonEmptyString(body('twoFactorContacts.*')),
-  nonEmptyString(body('notes')).optional(),
+  body('notes').optional(),
```
chpy04: shouldn't this be .isString()?

src/backend/src/services/finance.services.ts
```diff
@@ -1096,15 +1096,17 @@ export default class FinanceServices {
...
-    const existingSponsor = await prisma.sponsor.findFirst({
-      where: {
-        name,
-        organizationId: organization.organizationId
-      }
-    });
+    if (name !== oldSponsor.name) {
+      const existingSponsor = await prisma.sponsor.findFirst({
+        where: {
+          name,
```
chpy04: can we do case insensitive check here as well

chpy04 (changes requested):

PR #3771, opened 2025-11-19
#3706 your events page

### LINKED ISSUE #3706: [Calendar Improvements] - your events page
### Description

We need a your events page under the tab to show the events the current user has made. You should copy the table from another page, so it is cohesive. No need to worry about the warnings.

### Acceptance Criteria

Shows a spreadsheet all of the users events
Events that have to be approved by another member show the status of the approval
Events that do not have to be approved by another show N/A for Approval By and Approval Status

### Mocks

Image

Santiordon:
## Changes

Implementation of "Your Events" page. Composed of a simple page showing the events of the specified user.

## Notes

There is one slight concern if there is going to be a push for expanding to more platforms, there is not much security for the Filtering events endpoint. Anyone can get any sort of event that anyone is registered to. If there is a privileged event, anyone can fetch that event as of now.

Nothing else to note.

## Test Cases

- Case A
- Not approved / Approved "Approved By" category
- ...

## Screenshots

yourEventsSmall
yourEventsBig

None. API Endpoints already implemented.

## To Do

None

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3706 (issue 3706)


src/frontend/src/pages/NewCalendarPage/YourEventsPage.tsx
```diff
@@ -0,0 +1,202 @@
...
+      setEventsLoading(false);
+    });
+  }, [filterEventsMutate, user.userId]);
+
+  useEffect(() => {
+    const timer = setInterval(() => {
+      setUpdate((prev) => !prev);
+    }, 1000);
+    return () => clearInterval(timer);
+  }, []);
```
JoshuaGoldberg, line 83: What is the Effect for? I'm assuming it helps re-render the time until down below, which if so I think is fine?

src/frontend/src/pages/NewCalendarPage/YourEventsPage.tsx
```diff
@@ -0,0 +1,202 @@
...
+    const filterArgs: FilterArgs = {
+      memberIds: [user.userId],
+      startPeriod: new Date(0),
+      endPeriod: new Date(2099, 11, 31) // Adjust as needed
+    };
+    filterEventsMutate(filterArgs).then((events) => {
+      setEvents(events);
+      setEventsLoading(false);
+    });
+  }, [filterEventsMutate, user.userId]);
```
JoshuaGoldberg: This Effect seems a bit unnecessary to me. I think you can replace
`const { mutateAsync: filterEventsMutate } = useFilterEvents();` with
`const { data: events, isLoading } = useFilterEvents( *filter args* );`
and then slightly modify the hook to get a nicer representation (then you can plug in isLoading instead of defining a new state, same with events). If you have any questions about it/this doesn't actually work, do let me know!

src/frontend/src/hooks/calendar.hooks.ts
```diff
@@ -172,3 +174,18 @@ export const useDeleteMachinery = () => {
     }
   );
 };
+
+export const useFilterEvents = () => {
+  const qc = useQueryClient();
+  return useMutation<Event[], Error, FilterArgs>(
+    async (payload) => {
```
JoshuaGoldberg: Instead of taking in a payload via the async function, I think it might be cleaner to write it as a useQuery instead of useMutation, similar to
```export const useFilterEvents = (filterArgs: FilterArgs) => {
  return useQuery<Event[], Error>(['filter-events'], async () => {
    const { data } = await postFilterEvents(filterArgs);
    return data;
  });
};
```
as this (should) allow for a simpler way to grab all filtered events

src/frontend/src/apis/transformers/calendar.transformer.ts
```diff
@@ -8,3 +8,15 @@ export const shopTransformer = (shop: Shop): Shop => {
     userCreated: userTransformer(shop.userCreated)
   };
 };
+
+export const filterEventsTransformer = (events: Event[]): any => {
```
JoshuaGoldberg: This should probably just be an filterEventTransformer for individual events (in case we need to use it later), and you can just call map in the api

JoshuaGoldberg (changes requested):
This looks great! Left some comments mainly on design choices, but cool stuff overall 👍

src/frontend/src/pages/NewCalendarPage/YourEventsPage.tsx
Santiordon: Yep! It just makes sure that the countdown is refreshed since I didn't want to separate the cells into a different React Component (pretty redundant). It just guarantees that the page is refreshed each second. In terms of performance it would be a problem if this was a pretty packed page but all it is is just some cells so it doesn't really impact much performance.

src/frontend/src/pages/NewCalendarPage/YourEventsPage.tsx
Santiordon: Hm... Interesting. I will look into experimenting with that.

src/frontend/src/hooks/calendar.hooks.ts
Santiordon: That does look cleaner! I will definitely change it to that.

src/frontend/src/apis/transformers/calendar.transformer.ts
Santiordon: Good point. Will split that up!

PR #3779, opened 2025-11-21
#3759 update active ruleset

### LINKED ISSUE #3759: [RulesDashboard] - updateActiveRuleset Endpoint
### Description

We need an endpoint to update the active status of a ruleset so that users can simply check/uncheck a box on the frontend to select the active ruleset. This will require a new route, controller function, service function, and tests. The route should be `rules/rulesetType/:rulesetTypeId/:rulesetId/update`

### Acceptance Criteria

- [ ] There are new functions to accommodate for updating the status of a ruleset (this can more or less be seen as the editRuleset endpoint, but status is the only thing we need to edit as far as the feature goes)

### Proposed Solution

Implement new functions in the corresponding files and add tests. Might also want to consider adding a new field + unique constraint on activeRuleset to enforce that only one ruleset can be active at once.

Try to look around the codebase and see if there are any other places where we handle a similar situation, since I'm curious as to whether or not a unique constraint is the move

### Mocks

How this endpoint will be used on the frontend, as reference

Image
Image

staysgt:
## Changes

Created update ruleset status endpoint

## Test Cases

- successful update
- updating status to true when there is already an active ruleset fails
- fails when non-leadership attempts to update

## Screenshots
Screenshot 2025-11-25 at 6 54 04 PM

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3759


src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,110 @@ export default class RulesService {
...
+    return projectRuleTransformer(deletedProjectRule);
+  }
+
+  static async updateActiveRuleset(submitter: User, organizationId: string, rulesetId: string, status: boolean) {
+    if (!(await userHasPermission(submitter.userId, organizationId, isLeadership))) {
+      throw new AccessDeniedException('You do not have permissions to update ruleset status');
+    }
+
+    if (status) {
+      const activeRuleset = await prisma.ruleset.findFirst({
```
cielbellerose, line 755: make sure this is looking for another active ruleset in this specific ruleset type, so we should add a check to make sure its rulesetTypeId is the same as the current ruleset's

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,110 @@ export default class RulesService {
...
+    });
+
+    return projectRuleTransformer(deletedProjectRule);
+  }
+
+  static async updateActiveRuleset(submitter: User, organizationId: string, rulesetId: string, status: boolean) {
+    if (!(await userHasPermission(submitter.userId, organizationId, isLeadership))) {
+      throw new AccessDeniedException('You do not have permissions to update ruleset status');
+    }
+
```
cielbellerose, line 739: also make sure the current ruleset being updated exists and is not deleted, and that it is in the correct org

src/backend/tests/unit/rule.test.ts
```diff
@@ -385,6 +385,32 @@ describe('Create Rules Tests', () => {
...
+    it('update ruleset status - fails if one is already active', async () => {
+      const ruleset2 = await RulesService.createRuleset(
+        superman,
+        organization,
+        'ruleset name',
+        rulesetType.rulesetTypeId,
+        0,
+        false,
+        'fileId'
+      );
```
cielbellerose, line 489: we should also test
- succeeds even if an active ruleset exists in another ruleset type
- fails if wrong org ruleset or if ruleset is deleted
- succeeds setting an active ruleset to inactive (false value input)

cielbellerose (changes requested):
great work so far! just a few comments

cielbellerose (approved):
lgtm! just needs docstring on service function

src/backend/src/controllers/rules.controllers.ts
```diff
@@ -137,4 +147,22 @@ export default class RulesController {
       next(error);
     }
   }
+
+  static async updateActiveRuleset(req: Request, res: Response, next: NextFunction) {
```
chpy04: updateActiveRuleset is isleading because this can update any ruleset

src/backend/src/controllers/rules.controllers.ts
```diff
@@ -137,4 +147,22 @@ export default class RulesController {
...
+    try {
+      const { rulesetId } = req.params;
+      const { status } = req.body;
```
chpy04: woudn't we also want to edit the name here?

src/backend/src/routes/rules.routes.ts
```diff
@@ -54,4 +55,11 @@ rulesRouter.post(
 );
 rulesRouter.post('/rulesetType/:rulesetTypeId/delete', RulesController.deleteRulesetType);
 
+rulesRouter.post(
+  '/ruleset/:rulesetId/update',
+  body('status').isBoolean(),
```
chpy04: rename to isActive because status ussually implies an enum

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,139 @@ export default class RulesService {
...
+  /**
+   * Deletes a project rule and its associated rule status changes
+   * @param projectRuleId The ID of the project rule to delete
+   * @param deleter The user deleting the project rule (must be admin)
+   * @param organization The organization the project rule belongs to
+   * @returns The deleted project rule
+   */
+  static async deleteProjectRule(projectRuleId: string, deleter: User, organization: Organization): Promise<ProjectRule> {
+    if (!(await userHasPermission(deleter.userId, organization.organizationId, isAdmin))) {
```
chpy04: heads or lead / manager on the project

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,139 @@ export default class RulesService {
...
+    if (projectRule.rule.ruleset.car.wbsElement.organizationId !== organization.organizationId) {
+      throw new InvalidOrganizationException('Project Rule');
+    }
+
+    if (projectRule.dateDeleted) {
+      throw new DeletedException('Project Rule', projectRuleId);
+    }
+
+    const deletedProjectRule = await prisma.$transaction(async (tx) => {
+      await tx.rule_Status_Change.updateMany({
```
chpy04: the rule_Status_Change table is used for history, so we never want to delete from it

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,139 @@ export default class RulesService {
...
+    if (!(await userHasPermission(submitter.userId, organizationId, isLeadership))) {
+      throw new AccessDeniedException('You do not have permissions to update ruleset status');
+    }
+
+    const rulesetExists = await prisma.ruleset.findUnique({
+      where: {
+        rulesetId,
+        rulesetType: {
+          organizationId,
+          deletedByUserId: null
```
chpy04: in general we only care about whether or not object we are looking at is deleted or not. Even though this should always be the case, it could get very confusing for someone if they are getting a ruleset not found error but the issue is actually the type, so I would prefer to remove this check

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,139 @@ export default class RulesService {
...
+        },
+        deletedByUserId: null
```
chpy04, line 746: this will make your deleted check below never hit because the response is already filtered so you can remove this check

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,139 @@ export default class RulesService {
...
+      }
+    });
+
+    if (!rulesetExists) {
+      throw new NotFoundException('Ruleset', rulesetId);
+    }
+
+    if (rulesetExists.deletedByUserId) {
+      throw new NotFoundException('Ruleset', rulesetId);
```
chpy04: DeletedException

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,139 @@ export default class RulesService {
...
+    }
+
+    if (status) {
```
chpy04: just to prevent a situation where we add other fields to this endpoint (like name) and someone passes in the status as true when the ruleset is already active, can you make this if (!rulesetExists.active && status)

src/backend/src/services/rules.services.ts
```diff
@@ -568,4 +568,139 @@ export default class RulesService {
...
+          active: true,
+          rulesetType: {
+            rulesetTypeId: rulesetExists.rulesetTypeId,
+            organizationId
+          },
+          dateDeleted: null
+        }
+      });
+
+      if (activeRuleset && activeRuleset.rulesetId !== rulesetId) {
```
chpy04: if you do the above check this check is redundant

chpy04 (changes requested):

src/backend/src/services/rules.services.ts
```diff
@@ -702,23 +712,76 @@ export default class RulesService {
...
+
+  /**
+   * Updates a rulesets status
+   * @param submitter user updating the ruleset
+   * @param organizationId organization of ruleset being updated
+   * @param rulesetId id of ruleset being updated
+   * @param status new status of ruleset
+   * @returns
+   */
+  static async updateRuleset(submitter: User, organizationId: string, rulesetId: string, name: string, status: boolean) {
```
chpy04: rename status to isActive

chpy04 (changes requested):

chpy04 (approved):

PR #3783, opened 2025-11-23
#3754 migration to make design reviews and meetings event types

### LINKED ISSUE #3754: [Calendar Improvements] - migration to transform design reviews and meetings into event types/events without data loss
### Description

We need to get rid of design reviews and meetings table and make them into event types and events without losing data.

wavehassman:
## Changes

An insane amount. Made Design Reviews and Meetings event types and events. Everything still works except editing and creating modals, but everything else should be good.

Very crude changes to things because a lot of it will be rewritten with the new functionality.

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3754


src/backend/src/prisma/migrations/20251101233144_calendar/migration.sql
```diff
@@ -362,3 +357,328 @@ ALTER TABLE "public"."_CalendarToEvent_Type" ADD CONSTRAINT "_CalendarToEvent_Ty
...
+        WHEN 2 THEN 'TUESDAY'::public."DayOfWeek"
+        WHEN 3 THEN 'WEDNESDAY'::public."DayOfWeek"
+        WHEN 4 THEN 'THURSDAY'::public."DayOfWeek"
+        WHEN 5 THEN 'FRIDAY'::public."DayOfWeek"
+        WHEN 6 THEN 'SATURDAY'::public."DayOfWeek"
+    END],
+    dr."dateScheduled" + ((10 + time_slot) * INTERVAL '1 hour'), -- 10am + time_slot
+    dr."dateScheduled" + ((11 + time_slot) * INTERVAL '1 hour'), -- End time (1 hour later)
+    0, -- No recurrence for design reviews
+    dr."dateScheduled", -- initialDateScheduled is the actual scheduled date
```
chpy04: should we use the design review's initialDateScheduled?

src/backend/src/prisma/migrations/20251101233144_calendar/migration.sql
```diff
@@ -362,3 +357,328 @@ ALTER TABLE "public"."_CalendarToEvent_Type" ADD CONSTRAINT "_CalendarToEvent_Ty
...
+-- Link Schedule_Slots to Events for Design Reviews
+INSERT INTO "public"."_EventToSchedule_Slot" ("A", "B")
+SELECT 
+    dr."designReviewId",
+    ss."scheduleSlotId"
+FROM "public"."Design_Review" dr
+JOIN "public"."Schedule_Slot" ss ON 
+    ss."initialDateScheduled" = dr."dateScheduled"
+    AND ss."endDate" = dr."dateScheduled"
+    AND DATE_PART('hour', ss."startTime") - 10 = ANY(dr."meetingTimes");
```
chpy04: I don't think this would be an issue, but if two DRs had the same schedule, I think they would both get added to both schedule slots by this code, so wanted to make sure you are aware of this behavior

src/backend/src/prisma/migrations/20251101233144_calendar/migration.sql
```diff
@@ -362,3 +357,328 @@ ALTER TABLE "public"."_CalendarToEvent_Type" ADD CONSTRAINT "_CalendarToEvent_Ty
...
+CROSS JOIN LATERAL unnest(m."meetingTimes") AS time_slot;
+
+-- Link Schedule_Slots to Events for Meetings
+INSERT INTO "public"."_EventToSchedule_Slot" ("A", "B")
+SELECT 
+    m."meetingId",
+    ss."scheduleSlotId"
+FROM "public"."Meeting" m
+JOIN "public"."Schedule_Slot" ss ON 
+    ss."initialDateScheduled" = m."dateSet"::DATE
```
chpy04: same thing here

src/backend/src/prisma/schema.prisma
```diff
@@ -1100,12 +1091,11 @@ model Event_Type {
   userDeletedId                    String?
   userDeleted                      User?        @relation(fields: [userDeletedId], references: [userId], name: "eventTypeDeleter")
   calendars                        Calendar[]
-  initialDateScheduled             Boolean
-  allDay                           Boolean
-  recurring                        Boolean
+  schedule                         Boolean
```
chpy04: shouldn't everything have a schedule? I feel like this should be required for every event type

src/backend/src/prisma/migrations/20251101233144_calendar/migration.sql
```diff
@@ -362,3 +357,328 @@ ALTER TABLE "public"."_CalendarToEvent_Type" ADD CONSTRAINT "_CalendarToEvent_Ty
...
+REFERENCES "Event"("eventId")
+ON DELETE SET NULL   
+ON UPDATE CASCADE;
+
+-- Drop old relation tables for Design_Review
+DROP TABLE IF EXISTS "public"."_requiredAttendee" CASCADE;
+DROP TABLE IF EXISTS "public"."_optionalAttendee" CASCADE;
+DROP TABLE IF EXISTS "public"."_confirmedAttendee" CASCADE;
+DROP TABLE IF EXISTS "public"."_deniedAttendee" CASCADE;
+DROP TABLE IF EXISTS "public"."_userAttended" CASCADE;
```
chpy04: I'd prefer to not do the IF EXISTS because these should exist and if they don't I would rather it error to say that something went wrong

src/backend/src/prisma/migrations/20251101233144_calendar/migration.sql
```diff
@@ -362,3 +357,328 @@ ALTER TABLE "public"."_CalendarToEvent_Type" ADD CONSTRAINT "_CalendarToEvent_Ty
...
+
+-- Drop the old Meeting and Design_Review tables
+DROP TABLE IF EXISTS "public"."Meeting" CASCADE;
+DROP TABLE IF EXISTS "public"."Design_Review" CASCADE;
```
chpy04: same

src/backend/src/prisma/schema.prisma
```diff
@@ -515,7 +507,6 @@ model WBS_Element {
   assemblies                  Assembly[]
   materials                   Material[]
   reimbursementProductReasons Reimbursement_Product_Reason[]
-  designReviews               Design_Review[]
```
chpy04, line 518: we need to link these to events, I thought we were already doing that but if not we have to add it

src/backend/src/services/calendar.services.ts
```diff
@@ -447,6 +469,69 @@ export default class CalendarService {
       ...getEventQueryArgs(organization.organizationId)
     });
 
+    if (workPackageIds.length > 0) {
```
chpy04: I feel like it would be confusing to users that events with work packages are treated like design reviews and have slack notifs where other events don't. Would it be possible to just have slack notifications as a boolean field on the event type, where if it is true it notifies users (using the names to populate the messages instead of assuming its a DR) as well as the wp project slack channels? I think more generally I either want to completely get rid of the DR specific functionality (which it looks like we are very close to) or add the inheritance structure with a DR model in the schema that I mentioned earlier. LMK thoughts on this though I'm open to other ideas, my main concern is users won't understand when notifications happen

src/backend/src/services/calendar.services.ts
```diff
@@ -794,7 +896,13 @@ export default class CalendarService {
       });
     });
 
-    return eventTransformer(updatedEvent);
+    const edittedEvent = eventTransformer(updatedEvent);
+
+    if (status === Event_Status.SCHEDULED && workPackageIds.length > 0) {
```
chpy04: same thing here with the slack notifs

src/backend/src/services/calendar.services.ts
```diff
@@ -903,8 +1011,9 @@ export default class CalendarService {
         }
       });
 
-      // possibly want to do
-      //await sendDRUserConfirmationToThread(updatedDesignReview.notificationSlackThreads, submitter);
+      if (updatedEvent.workPackages.length > 0) {
```
chpy04: and here

src/backend/src/services/calendar.services.ts
```diff
@@ -918,8 +1027,9 @@ export default class CalendarService {
             status: Event_Status.CONFIRMED
           }
         });
-
-        //await sendDRConfirmationToThread(updatedDesignReview.notificationSlackThreads, updatedDesignReview.userCreated);
+        if (updatedEvent.workPackages.length > 0) {
```
chpy04: and here, also thinking about the confirmation messages, we might also need a "requireConfirmation" boolean to determine that in the team types, but you've been working with these models more so what do you think

src/backend/src/services/notifications.services.ts
```diff
@@ -119,70 +120,94 @@ export default class NotificationsService {
   /**
    * Sends the design review slack notifications for all design reviews scheduled for today
    */
-  static async sendDesignReviewSlackNotifications() {
+  static async sendDesignReviewEventSlackNotifications() {
```
chpy04: same thing here about differentiating between DRs and other

src/backend/src/utils/google-integration.utils.ts
```diff
@@ -258,36 +263,47 @@ export const createCalendarEvent = async (
...
   zoomLink: string | null,
   location: string | null,
-  meetingTimes: number[],
-  wbsElement: WBS_Element
+  eventTitle: string
 ) => {
+  const [firstSlot] = scheduledSlots;
+  if (!firstSlot?.startTime || !firstSlot?.endTime) {
+    throw new Error('Event must have a valid start and end time');
+  }
```
chpy04, line 287: we would need to validate all the slots and add all of them not just hte first one, but also the google calendar integration isn't even used rn so I'm fine just commenting this out and making a future ticket

src/backend/src/utils/pop-up.utils.ts
```diff
@@ -1,31 +1,31 @@
...
-  designReview: Design_Review,
+export const sendDrEventPopUp = async (
+  event: Event,
   members: User[],
   submitter: User,
   workPackageName: string,
   organizationId: string
 ) => {
-  const designReviewLink = `/settings/preferences?drId=${designReview.designReviewId}`;
+  const designReviewEventLink = `/settings/preferences?drId=${event.eventId}`;
```
chpy04: does this need to be eventId instead of drId? It looks like its changed below but not here

chpy04 (changes requested):

src/backend/src/services/calendar.services.ts
```diff
@@ -2038,4 +2145,38 @@ export default class CalendarService {
...
+    return eventTransformer(event);
+  }
+
+  /**
+   * Gets all events in the database
+   * @param organizationId the organization id of the current user
+   * @returns All of the events
+   */
+  static async getAllEvents(organization: Organization): Promise<Event[]> {
+    const events = await prisma.event.findMany({
```
JoshuaGoldberg: is getAllEvents (and get single event) needed here? Or could the controller simply be a wrapper for FilterEvents that passes in the necessary args (None for getAll and the eventID for get single event)

src/backend/src/prisma/schema.prisma
wavehassman: I agree, but it's on the event type modal, so do we want to take it off and just make it required?

src/backend/src/services/calendar.services.ts
wavehassman: I'm thinking we should just check that the event status is confirmed before allowing a user to confirm their availability but lmk

src/backend/src/utils/google-integration.utils.ts
wavehassman: So we still want the google integration even with the calendar improvements? Isn't the point to make the gcal unecessary

src/backend/src/prisma/schema.prisma
wavehassman: Right now, events are linked to work packages and I think for the modal that's good like just choose a work package, but then in the schema we link to the Wbs_Element, this kinda ruins a lot of my changes so lmk what you think before I go ahead on this

chpy04: oh ok I missed that this is fine then

src/backend/src/services/calendar.services.ts
chpy04: Confirmed would be after everyone confirms their availability, but I was more saying I don't think every meeting should go through the whole scheduling process that DRs go through. Like most events people are just going to schedule it and the people that make it make it, so we need some way to bypass the confirmation step on certain meetings

src/backend/src/utils/google-integration.utils.ts
chpy04: Yeah I'm just saying it would be nice to have a g cal too bc ppl like to have their own personal calendars but that is def out of scope for this feature

wavehassman: ok i'll make it a ticket

src/backend/src/services/calendar.services.ts
wavehassman: When creating an event, it checks the requires confirmation boolean in the event type. I made it so when creating an event if it is automatically confirmed, it will send the confirmation message, and if it is edited to confirmed then it will send as well

chpy04: perfect

src/backend/src/prisma/schema.prisma
chpy04: yes thats def something left in the epic by accident, if someone made an event without a schedule there would be no way to even see it. Doesn't have to be this ticket though you can do that as a seperate ticket

src/backend/src/services/calendar.services.ts
```diff
@@ -69,6 +76,7 @@ export default class CalendarService {
    * @param documents Determines if documents are associates with this event type.
    * @param description Determines if a description is associated with this event type.
    * @param onlyHeadsOrAbove Determines if events under this event type can only be created by heads or above.
```
chpy04, line 77: missing requiresConfirmation

chpy04 (approved):

PR #3878, opened 2026-01-15
#3867 schema changes

### LINKED ISSUE #3867: [BOM Improvements] - Schema Changes
### Description

make manufacturer, all part details, and all price information optional

Reimbursement products should be relations to BOMs

wavehassman:
## Changes

schema changes to make manufacturer, part details and price information optional
made relation between materials and reimbursement product
changed types and endpoints to match these changes

## Checklist

- [x] No merge conflicts
- [x] All checks passing

Closes #3867


src/backend/src/prisma/migrations/20260115203541_bom_improvements/migration.sql
```diff
@@ -0,0 +1,30 @@
+/*
```
chpy04: We will need to migrate existing connections. Specifically, if a material is already connected to an RR, We should connect it to all the line items in that RR. I'm not sure if this is the case, but if there are multiple materials currently connected to a single RR, this will get more complicated. I think it would make sense to either make it a many to many relationship between materials and products so you can just add them all to the first, or you could just add reimbursement products to the previous RRs with a cost of 0. It might be worth asking engineers if they would every want a single product in an RR to cover multiple materials. My assumption would be no in that they should really split it up into two products each connected to their own material, but I don't know if theres some case where they would order one item that fullfills multiple materials

src/backend/src/prisma/schema.prisma
```diff
@@ -736,6 +736,8 @@ model Reimbursement_Product {
   name                         String
   dateDeleted                  DateTime?
   cost                         Int
+  material                     Material?                    @relation(fields: [materialId], references: [materialId])
+  materialId                   String?                      @unique
```
chpy04: One thing to potentially consider is that materials have quantities, so it is possible that someone would order some amount at one time and the rest at a different time. This would violate the unique constraint set here. Even if we don't get to that ability on the frontend this semester, I feel likes its worth leaving the unique constraint out in case we have time to get to it

chpy04 (approved):
Couple things to think about down the line but if you want to get this out for developers go for it

chpy04 (approved):
lgtm

PR #3977, opened 2026-02-17
#3887 Material Autocomplete

### LINKED ISSUE #3887: [BOM Improvements] - Replace RR Product Text Input with Material Autocomplete
### Description

In the Reimbursement Request form, when a project is selected, replace the "Product Name/Description" text input with an autocomplete dropdown that shows materials from the selected project. 

### Acceptance Criteria

- When no category is selected: existing product name text input appears 
- When a non-project category is selected: existing product name text input appears 
- When a project is selected: text input is replaced with autocomplete dropdown
- Autocomplete shows all materials from the selected project displaying: Material Name, Manufacturer, Material Type 
- User can search/filter materials by typing (search on name and part number)
- When a material is selected, it's stored as the product for that reimbursement item
- Multiple products can be added, each with material selection
- Form validation requires material selection for project-based products

nunnyu:
## Changes

When you create a reimbursement request, for "Purchased Products", when you select a project, rather than it being a vague description, there is an autocomplete dropdown that contains all materials linked to that project. The other categories still just do the basic description variant with no drop down. 

## Screenshots

image

image

image

#
Closes #3887


src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
```diff
@@ -62,6 +64,46 @@ const ListItem = styled('li')(({ theme }) => ({
...
+    return <LoadingIndicator />;
+  }
+
+  if (isError) {
+    return <ErrorPage message={error?.message || 'Failed to load materials'} />;
+  }
+
+  const materialOptions = (materials || []).map((material) => ({
+    id: material.materialId,
+    label: `${material.name}: ${material.materialTypeName}, ${material.manufacturerName}`
```
wavehassman: we want to be able to search on name and part number, so this should include the manufacturer part number. I would take out the material type name, and make the order name, manufacturer name, manufacturer part number

wavehassman (changes requested):
Looks awesome! Just a small change

rbessin:
Made a quick mistake with commit name and then made it worse! The code should work however. Sorry about the confusion.

src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
```diff
@@ -62,6 +64,51 @@ const ListItem = styled('li')(({ theme }) => ({
...
+
+  if (isLoading) {
+    return <LoadingIndicator />;
+  }
+
+  if (isError) {
+    return <ErrorPage message={error?.message || 'Failed to load materials'} />;
+  }
+
+  const materialOptions = (materials || []).map((material) => ({
```
chpy04: really nitpicky but also kind of useful for dev experience: if in the loading check you do 
`if (isLoading || !materials) ...`
the typing will guarentee that you have the materials object (which you already knew anyways), and you won't have to do the materials || []

src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
```diff
@@ -62,6 +64,51 @@ const ListItem = styled('li')(({ theme }) => ({
   margin: theme.spacing(0.5)
 }));
 
+const MaterialAutocomplete: React.FC<{
+  wbsNum: WbsNumber;
+  onSelect: (material: Material) => void;
+  initialValue?: string;
+}> = ({ wbsNum, onSelect, initialValue }) => {
+  const { data: materials, isLoading, isError, error } = useGetMaterialsForWbsElement(wbsNum);
+  const [userSelected, setUserSelected] = useState<{ id: string; label: string } | null>(null);
```
chpy04: materialSelected and setMaterialSelected instead of user right?

chpy04 (changes requested):

src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
```diff
@@ -62,6 +64,51 @@ const ListItem = styled('li')(({ theme }) => ({
...
+    return <LoadingIndicator />;
+  }
+
+  if (isError) {
+    return <ErrorPage message={error?.message || 'Failed to load materials'} />;
+  }
+
+  const materialOptions = (materials || []).map((material) => ({
+    id: material.materialId,
+    label: `${material.name}: ${material.manufacturerName}, ${material.manufacturerPartNumber}`
```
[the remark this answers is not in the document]
chpy04: I agree here. I know Waverly was planning on adding some fuzzy search implementations in the future, but at least for now we should include this data in the material strings so that users can search over it within a BOM. just make sure the text is reasonably formatted

src/frontend/src/pages/FinancePage/ReimbursementRequestForm/ReimbursementProductTable.tsx
```diff
@@ -62,6 +64,51 @@ const ListItem = styled('li')(({ theme }) => ({
...
+}> = ({ wbsNum, onSelect, initialValue }) => {
+  const { data: materials, isLoading, isError, error } = useGetMaterialsForWbsElement(wbsNum);
+  const [userSelected, setUserSelected] = useState<{ id: string; label: string } | null>(null);
+
+  if (isLoading) {
+    return <LoadingIndicator />;
+  }
+
+  if (isError) {
+    return <ErrorPage message={error?.message || 'Failed to load materials'} />;
```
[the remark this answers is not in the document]
chpy04: idrc about this but if you want to copilot is prolly right🤷‍♂️

rbessin:
The reason why we were getting the bug when creating a new request but not when editing was because they have two different endpoints; create forces the body to include a materialId whereas edit allows either the name or the materialId to be passed. Since materialId was not being passed as a property when the code built the WBS reimbursement product objects, the backend ended up rejecting the requests every time when creating new reimbursement requests but not when editing them. Is it intended for the two endpoints to have different requirements? Shouldn't we enforce the edit to include a materialId? It's probably not possible to make this change since we would have to modify all existing data but I just wanted to point out this fact.

wavehassman:
> The reason why we were getting the bug when creating a new request but not when editing was because they have two different endpoints; create forces the body to include a materialId whereas edit allows either the name or the materialId to be passed. Since materialId was not being passed as a property when the code built the WBS reimbursement product objects, the backend ended up rejecting the requests every time when creating new reimbursement requests but not when editing them. Is it intended for the two endpoints to have different requirements? Shouldn't we enforce the edit to include a materialId? It's probably not possible to make this change since we would have to modify all existing data but I just wanted to point out this fact.

Yes, the endpoints needs two different requirements because we only want reimbursement requests to be created with a materialId when it is on a project from this point forward. But, there are reimbursement requests that exist in the db that still just hold the name as a string, so we need to be able to handle both cases. You should make the create endpoint and edit endpoint work under these conditions. I have changed the schema to include materialId optionally, so you should only need to change the endpoints and potentially frontend types that I missed. For this ticket, you need to make sure the edit and view modals work for both string and material based data, but create only needs to work for material based data when creating a reimbursement request on a project.

chpy04 (approved):
lgtm!

wavehassman (approved):
amazing!

PR #4005, opened 2026-03-03
#3948 update get work packages endpoints

### LINKED ISSUE #3948: [Maintenance Car] - Update getWorkPackages Endpoints
### Description

Update the used getWorkPackages endpoints. This would be getManyWorkPackages and getAllWorkPackages endpoints. Include an optional car parameter to filter for only packages of that car id.

### Acceptance Criteria

The getManyWorkPackages endpoint and any other endpoint that requires it support an optional car parameter that returns only those work packages.

patriots1:
## Changes

Added optional car parameter for filtering to getAllWorkPackages, getManyWorkPackages, getBlockingWorkPackages and getHomePageWorkPackages.

## Test Cases

Unable to test due to issues with local machine. Unit test file for work-packages did not exist. 

## To Do

- [ ] test updated endpoints

Closes #3948


Steph375 (commented):
I just added middleware that sets req.currentCar based on the carId header from the global car filter. Because of that, controllers shouldn't read carId from the request body anymore.

This endpoint can use req.currentCar?.carId instead so it stays consistent with the rest of the controllers.   I would look at getAllProjects its doing something similar.

patriots1:
Refactored controller to match middleware. Service logic was buggy due to changing the type of carId - fixed.

PR #4017, opened 2026-03-07
Prisma config update

chpy04:
## Changes

Updated prisma configuration to no longer use deprecated URL in schema syntax


chpy04:
Idk why the connection pooling isn't working and this doesn't actually break anything so closing this for now

PR #4028, opened 2026-03-10
Feature/global car filtering

### LINKED ISSUE #3629: [Maintenance] - Global & Finance Filtering
### Description

1. FinishLine displays all data for all cars simultaneously, and users cannot filter the data on the entire website by car. This is not sustainable as the team grows and we build more cars.
2. Engineers do not currently understand the use of "Start Date" and "End Date" on the Finance Dashboard. Because of this, they do not filter the data on their Finance Dashboard by cars/timeframes. 

### Acceptance Criteria

1. Implement global car filter design
2. Implement car drop-down for all users to filter their Finance Dashboards

**Global Car Filter**
Allow users to filter data across all pages on FinishLine by entering the sidebar and selecting the car for which they want to see data. Notes on how users can use this filter:
- Can change car by clicking on the drop-down arrow, which will show all of the cars built in reverse chronological order. Once users select a new car, the drop-down should automatically close. 
- As more cars are added, the set of buttons with car names should have horizontal scroll functionality.
- Filter should default to current car and will only last for that session. 
- If the user changes the filtering on a specific page, then the data will only change on that page and the global filter will not change. If they leave that page after changing the filtering and return, it will default back to the global filter selection. Example: Global filter is set to 25A. User goes to Gantt Chart and filters data by 24A. User then leaves and visits Projects page. User comes back to Gantt Chart, which is filtered by 25A (global selection). 

**Car Filter for All Users on Finance Dashboard**
Implement car-specific filtering for all users on their Finance Dashboards (not just for Finance Team/Admin). Notes on additional functionality: 
- Since engineers don’t currently understand the functionality of start and end, allow them to select a car from the car drop-down. If they select a car, it will populate the start date to show when that car was initialized and the end date as today’s date (if current car) or the end date of that car (if previous car). They can then adjust these dates to account for non-car data, but it will provide clarity regarding which time frames this data is shown for. This should be implemented for all users. 
- A question mark hover next to the start and end dates to explain that they are used to filter for car data + non-car/category data (ex. comps, food, etc).

### Mocks

**Global Car Filter Design**
Image

cielbellerose:
## Changes

Pages updated for filtering included: Home, Gantt, Projects, Change Requests, Finance.
Pages excluded from global filtering: Teams, Statistics, Calendar, Restrospective.

- Add global car filter in sidebar.
- Add override capabilities for specific pages to override global filter.
- Add middleware to pass in car selection.
- Add backend filtering for endpoints linked to pages that should be filtered.

## Notes

Two features follow slightly different formats than the rest of the website for unique reasons.
- Gantt Chart overrides carId header to always capture all data; does front-end filtering; necessary due to local multi-car filtering.
- Finance pages take in global filter to display choice in page dropdowns; always flows choice back into the override.

Side changes that I believe were important to note.
- Moved seed data from Fergus to NER-25 to make tests simpler, since NER-25 is the most recent car.
- Fixed dropdowns to exclude guests (side change asked to be included in PR).

Compared to the ticket, some implementations are slightly different.
- Currently we default the car in the dropdown to the last globally selected car or all cars if not possible, however, the ticket indicates the default should always be the current car for the year.
- Currently the local filter on a page is preserved until the website is reloaded or until the next session, however, the ticket indicates the local filter should reset to global when the user navigates away from the page.

## Screenshots

image

## Checklist

- [X] No merge conflicts
- [X] All checks passing

Closes #3629


src/backend/src/controllers/finance.controllers.ts
```diff
@@ -169,20 +169,24 @@ export default class FinanceController {
     }
   }
 
+  // Finance data filters by carNumber (integer) rather than carId (UUID) because the
+  // finance schema links through WbsElement, which owns carNumber directly. Filtering
+  // by carId would require nested Prisma joins across every finance utility function.
+  // carNumber is sourced from req.currentCar?.wbsElement.carNumber via middleware.
```
chpy04: useful explanation but I think this comment should be inside the service function as opposed to a controller signature

src/backend/src/controllers/work-packages.controllers.ts
```diff
@@ -8,8 +8,15 @@ export default class WorkPackagesController {
   static async getAllWorkPackages(req: Request, res: Response, next: NextFunction) {
     try {
       const { query } = req;
+      const queryWithCarId = {
+        ...query,
+        carId: req.currentCar?.carId
```
chpy04: Is the idea here to override a carId field that is potentially in the query? If thats the case I would have thought it would have been the other way around where if its in the query the passed value would take precedence? Either way it's cleaner to just add a field to the service function. The patter of passing a query is depricated anyways its generally better to just extract and pass the arguments bc it makes the service function signature more straight forward

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -85,9 +85,22 @@ export default class ReimbursementRequestService {
...
+    organization: Organization,
+    carNumber?: number
+  ): Promise<ReimbursementRequest[]> {
     const userReimbursementRequests = await prisma.reimbursement_Request.findMany({
-      where: { dateDeleted: null, recipientId: recipient.userId, organizationId: organization.organizationId },
+      where: {
+        dateDeleted: null,
+        recipientId: recipient.userId,
+        organizationId: organization.organizationId,
+        ...(carNumber !== undefined && {
```
chpy04: This means we have to always guarantee that it will never be null. Probably is the case right now but just in terms of making it harder to break things down the line I always recommend doing `(carNumber !== undefined && carNumber !== null)` bc then you don't have to worry abt it

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -101,10 +114,20 @@ export default class ReimbursementRequestService {
...
+    organization: Organization,
+    carNumber?: number
   ): Promise<ReimbursementRequest[]> {
     const assignedReimbursementRequests = await prisma.reimbursement_Request.findMany({
-      where: { dateDeleted: null, assigneeId: assignee.userId, organizationId: organization.organizationId },
+      where: {
+        dateDeleted: null,
+        assigneeId: assignee.userId,
+        organizationId: organization.organizationId,
+        ...(carNumber !== undefined && {
```
chpy04: same

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -160,7 +184,12 @@ export default class ReimbursementRequestService {
       where: {
         dateDeleted: null,
         recipientId: { in: Array.from(teamUserIds) },
-        organizationId: organization.organizationId
+        organizationId: organization.organizationId,
+        ...(carNumber !== undefined && {
```
chpy04: same

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -593,7 +626,12 @@ export default class ReimbursementRequestService {
             type: Reimbursement_Status_Type.ADVISOR_APPROVED
           }
         },
-        accountCode: { organizationId: organization.organizationId }
+        accountCode: { organizationId: organization.organizationId },
+        ...(carNumber !== undefined && {
```
chpy04: same

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -980,13 +1018,25 @@ export default class ReimbursementRequestService {
...
     if (!(await isUserFinanceTeamOrHead(user, organization.organizationId))) {
       throw new AccessDeniedException(`You are not a member of the finance team!`);
     }
 
     const reimbursementRequests = await prisma.reimbursement_Request.findMany({
-      where: { dateDeleted: null, accountCode: { organizationId: organization.organizationId } },
+      where: {
+        dateDeleted: null,
+        accountCode: { organizationId: organization.organizationId },
+        ...(carNumber !== undefined && {
```
chpy04: same

src/backend/src/services/work-packages.services.ts
```diff
@@ -141,10 +149,15 @@ export default class WorkPackagesService {
...
    * @param organizationId the id of the organization that the user is currently in
+   * @param carId optional car number to filter work packages by
    * @returns the work packages with the given WBS numbers
    * @throws if any of the work packages are not found or are not part of the organization
    */
-  static async getManyWorkPackages(wbsNums: WbsNumber[], organization: Organization): Promise<WorkPackage[]> {
+  static async getManyWorkPackages(
+    wbsNums: WbsNumber[],
+    organization: Organization,
+    carId?: string
```
chpy04: Doesn't make sense to take a car to filter by here because the requester is already specifying which wbs numbers they want. Instead could you just filter the wbs nums arg by car wherever this is called from in the frontend? Also not entirely sure where this is used but it seems weird that we would want this to be sorted by car at all

src/backend/src/services/work-packages.services.ts
```diff
@@ -518,9 +544,14 @@ export default class WorkPackagesService {
...
    * @param wbsNum the wbs number of the work package to get the blocking work packages for
    * @param organizationId the id of the organization that the user is currently in
+   * @param carId the optional carId to filter work packages by
    * @returns the blocking work packages for the given work package
    */
-  static async getBlockingWorkPackages(wbsNum: WbsNumber, organization: Organization): Promise<WorkPackage[]> {
+  static async getBlockingWorkPackages(
+    wbsNum: WbsNumber,
+    organization: Organization,
+    carId?: string
```
chpy04: Shouldn't take a car here at all, every blocking work package should be on the same car and if its not I'd rather it have front end side effects than fail silently bc its filtered out here

src/frontend/src/app/AppGlobalCarFilterContext.tsx
```diff
@@ -0,0 +1,90 @@
...
+          setSelectedCarState(savedCar);
+          localStorage.setItem('selectedCarId', savedCar.id);
+          setIsInitialized(true);
+          return;
+        }
+        // Stored ID not found in car list (stale or invalid) — clear it
+        localStorage.removeItem('selectedCarId');
+      }
+
+      // Default to null (all cars)
```
chpy04: Can we make it so there's a distinct value for the local storage for when all cars are selected, and then if you can't find anything in local storage default to the most recent car? Trying to make sure people actually use the feature bc if we release it with defaulting to all no-one will know it exists

src/frontend/src/components/GlobalCarFilterDropdown.tsx
```diff
@@ -0,0 +1,121 @@
...
+              }}
+            />
+          );
+        })}
+      </Box>
+    </Box>
+  );
+};
+
+export default GlobalCarFilterDropdown;
```
chpy04, line 23: Amazing

src/frontend/src/hooks/finance-car-filter.hooks.ts
```diff
@@ -0,0 +1,101 @@
...
+ * directly as overrideCarId using selectedCar?.id ?? null, keeping query keys reactive to both.
+ *
+ * When a specific car is selected, dates auto-populate:
+ * - Start date: When the car was initialized (car.dateCreated)
+ * - End date: Today (if current car) or start date of the next car (if previous car)
+ */
+export const useFinanceDashboardCarFilter = (initialStartDate?: Date, initialEndDate?: Date): FinanceDashboardCarFilter => {
+  const { selectedCar: globalSelectedCar, allCars, isLoading, error } = useGlobalCarFilter();
+
+  // undefined = not set (mirror global), null = explicitly set to "All Cars", Car = explicitly set to specific car
```
chpy04: Should not have null and undefined have two different interpretations. Would rather have a union type with a string such as Car | "all cars". Also this might become more clear if we have an explicit value for all cars as described above

src/frontend/src/utils/axios.ts
```diff
@@ -37,6 +51,11 @@ axios.interceptors.request.use(
     if (import.meta.env.MODE === 'development') request.headers!['Authorization'] = localStorage.getItem('devUserId') || '';
     const organizationId = localStorage.getItem('organizationId');
     request.headers!['organizationId'] = organizationId ?? '';
+    if (request.overrideCarId !== undefined) {
+      if (request.overrideCarId !== null) request.headers!['carId'] = request.overrideCarId;
```
chpy04: same thing with null vs undefined, shouldn't have differing logic for these cases

chpy04 (changes requested):
Really good work, main thing is dedicated all cars handling instead of implicit with undefined. Also in terms of UI, on the selector:

Image

Can we
- add some spacing on to of the "Working with" so its not so close to the top of the screen
- Make the car options go all the way to the right below the sidebar arrow instead of getting horizontally cut off even when there is nothing to the right of them

chpy04 (approved):
bangggggggggggggggg

PR #4046, opened 2026-03-14
#3945 don't send slack notif when user is assignee

### LINKED ISSUE #3945: [Maintenance] - Remove Slack Notif Personal Task Assigned
### Description
If a user assigns a task to themself, we don't need to DM them on slack to inform them that they were assigned, since they already know

Typically, when a user is assigned a task on Finishline, they are sent a Finishline by NER slack DM informing them that they were assigned. 
Heres an example of what the message looks like just for clarity

Image

The goal here is to add a check so if the task creator matches the task assignee then don't send the slack ping.

### Proposed Solution

sendSlackTaskAssignedNotificationToUsers in backend/src/utils/tasks.utils.ts sends a task assigned notification to the specified users on Slack ..
- update this to also take in the id of who is assigning the task and ensure it doesn't match the assignee before sending the slack notif

### Acceptance Criteria

- [ ] if we assign a task to ourself, no slack dm is sent
- [ ] if we assign someone else to the task, the slack dm is sent as usual

contact steph and I when you get to testing so we can run through how to test slack dms

getheobald:
## Changes

Added a check for user id when sending a slack task assignment notification such that no notification is sent when assigning tasks to oneself

## Notes

- There are 2 backend tests failing but they're not related so I think that's not my circus not my monkeys?
- There is a typo in the params on this function but I didn't change it since I assume it's not in the scope of this ticket. But if it is I will happily fix the typo!!!

## Test Cases

- Manually tested assigning myself a task, and no notif was sent as expected
- Manually tested assigning another user a task and notif was sent as expected
- Manually tested reassigning a task from myself to another user and notif was sent as expected
- Manually tested reassigning a task from another user to myself and no notif was sent as expected

## Screenshots

Not sure if I need to include screenshots for manual e2e tests but here's the slack dm output
image

## Checklist

- [X] No merge conflicts
- [X] All checks passing

Closes #3945

### DIFF
--- a/src/backend/src/services/tasks.services.ts
+++ b/src/backend/src/services/tasks.services.ts
@@ -122,7 +122,8 @@ export default class TasksService {
 
     const newTask = taskTransformer(createdTask);
 
-    await sendSlackTaskAssignedNotificationToUsers(newTask, assignees, organization.organizationId, createdBy.userId);
+    const nonSelfAssigneeIds = assignees.filter((id) => id !== createdBy.userId);
+    await sendSlackTaskAssignedNotificationToUsers(newTask, nonSelfAssigneeIds, organization.organizationId);
 
     return newTask;
   }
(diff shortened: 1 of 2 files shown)


chpy04 (changes requested):
This is kind of nit picky, but I'd rather this check be in the service function than in the slack utils function. The slack utils functions should be responsible for formatting the slack message and wrapping the api calls, but it should be the service function's resposibility to determine whether or not the slack message should be sent.

getheobald:
No probs! Made the change and retested

chpy04 (approved):
lit

PR #4053, opened 2026-03-16
#3943 descriptive finance slack msg

### LINKED ISSUE #3943: [Maintenance] - Finance Slack Msg More Info
### Description

In finance Slack messages, add a quick description (totalCost, vendor, etc) so they are easily identifiable without having to check Finishline

### Original Request

"in the [#finance](https://nu-electric-racing.slack.com/archives/C02BLC9DP2A) channel, when you have a thread for a reimbursment, would it be possible to have a quick description of the reimbursment sent in thread (ammount, vendor etc) so you dont need to go into FL to figure out which reimbursement it is?"

### Proposed Solution

backend/src/utils/slack.utils.ts #sendReimbursementRequestCreatedNotificationAndCreateMessageInfo 
- update this function to add more information to the msg 

**potential example:** @{user} created a reimbursement request for $35.00 High Performance Battery Pack (ID#: 000)

### Mocks

_No response_

getheobald:
## Changes

Updated sendReimbursementRequestCreatedNotificationAndCreateMessageInfo in slack utils to send a more descriptive message in the slack notification that gets sent when a reimbursement request is created.

## Notes
Do they also want more info in the other notifications, like if a reimbursement request is rejected? I didn't change any of those but tbh they might want more details on those too.

## Test Cases
I just kinda made sure it worked. Didn't test edge cases cause not sure what those would be.

## Screenshots
image

## Checklist

- [X] No merge conflicts
- [X] All checks passing

Closes #3943

### DIFF
--- a/src/backend/src/utils/slack.utils.ts
+++ b/src/backend/src/utils/slack.utils.ts
@@ -141,7 +141,16 @@ export const sendReimbursementRequestCreatedNotificationAndCreateMessageInfo = a
 
   const reimbursementRequest = await prisma.reimbursement_Request.findUnique({
     where: { reimbursementRequestId: requestId },
-    ...getReimbursementRequestQueryArgs(organizationId)
+    select: {
+      identifier: true,
+      totalCost: true,
+      description: true,
+      vendor: {
+        select: {
+          name: true
+        }
+      }
+    }
   });
 
   if (!reimbursementRequest) throw new HttpException(500, 'Reimbursement request does not exist!');
@@ -173,7 +182,7 @@ export const sendReimbursementRequestCreatedNotificationAndCreateMessageInfo = a
   const { messageInfoId, channelId, timestamp } = createdMessageInfo;
 
   // send reimbursement request description in slack thread
-  if (description !== '') {
+  if (description) {
     await sendThreadResponse(
       [{ messageInfoId, channelId, timestamp, changeRequestId: null }],
       `Description: ${description}`
--- a/src/backend/src/utils/slack.utils.ts
+++ b/src/backend/src/utils/slack.utils.ts
@@ -31,7 +31,6 @@ import { Prisma } from '@prisma/client';
 import { userTransformer } from '../transformers/user.transformer.js';
 import { SlackRichTextBlock } from '../services/slack.services.js';
 import UsersService from '../services/users.services.js';
-import { getReimbursementRequestQueryArgs } from '../prisma-query-args/reimbursement-requests.query-args.js';
 
 interface SlackMessageThread {
   messageInfoId: string;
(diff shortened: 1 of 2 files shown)


src/backend/src/utils/slack.utils.ts
```diff
@@ -151,13 +162,23 @@ export const sendReimbursementRequestCreatedNotificationAndCreateMessageInfo = a
...
       reimbursementRequestId: requestId,
       channelId: messageInfo.channelId,
       timestamp: messageInfo.ts
     }
   });
+
+  const { messageInfoId, channelId, timestamp } = createdMessageInfo;
+
+  // send reimbursement request description in slack thread
+  if (description !== '') {
```
Steph375: this should probably be more defensive like if description was null or undefined it could pass

Steph375 (changes requested):
This looks great good job! Just that one small fix.

src/backend/src/utils/slack.utils.ts
getheobald: I checked the schema and it looks like description defaults to empty string so it should never be null or undefined which is why I wrote it like this but let me know if that's wrong!

StephSayegh: ok then its fine

Steph375 (dismissed):

src/backend/src/utils/slack.utils.ts
```diff
@@ -127,18 +128,28 @@ export const sendSlackTaskAssignedNotification = async (
...
-  requestIdentifier: number,
   submitterId: string,
   organizationId: string
 ): Promise<void> => {
   if (process.env.NODE_ENV !== 'production' && !DEV_TESTING_OVERRIDE) return; // don't send msgs unless in prod
 
-  const msg = `${await getUserSlackMentionOrName(submitterId)} created a reimbursement request (ID#: ${requestIdentifier}) 💲`;
+  const reimbursementRequest = await prisma.reimbursement_Request.findUnique({
+    where: { reimbursementRequestId: requestId },
+    ...getReimbursementRequestQueryArgs(organizationId)
```
chpy04: Because we don't have to retunr a shared type, including all the query args is fetching a bunch of data that this endpoint doesn't actually need. Instead, lets only include the fields that are actually required below to speed up this call

src/backend/src/utils/slack.utils.ts
chpy04: I actually agree with Steph here, even if it isn't optional right now, this is kind of the perfect situation to use string truthy falsey ness. If (description) would cover the current case and also prevent a bug if someone in the future were to make this optional and forget to check this line if that makes sense

chpy04 (changes requested):

chpy04 (approved):

PR #4103, opened 2026-03-31
#4013 fix all change request table crash

### LINKED ISSUE #4013: [Maintenance] - All Change Requests table crash
### Observed Behavior

Opening the All Change Requests table sometimes causes the page to fail to load with the error "undefined" is not valid JSON, preventing the table from rendering.

### Expected Behavior

Should render the table 

### Steps to Reproduce

1. Navigate to the Change Requests section in the application.

2. Click on All Change Requests from the menu.

3. add 1 filter of any kind, submit so the table is filtered

4. hit the x once to remove the filtering, hit the x again to remove the filter altogether, now refresh the page and the bug will appear indefinitely (unless if you inspect the page and go to console and paste `localStorage.removeItem('changeRequestsTableFilter')` hit enter and then refresh) 

5. If an invalid filter value is stored in localStorage under changeRequestsTableFilter, the page attempts to parse it on load.

6. The page fails to render and the console shows the error "undefined" is not valid JSON.

### Screenshots (as needed)

_No response_

patriots1:
## Changes

When an item is removed from localStorage and it becomes undefined, we still access it using localStorage.setIterm. This is the main cause of the issues. I fixed this via adding a condition to check against undefined and added a try and catch statement to JSON.parse to catch any corrupted values in localStorage.

## Test Cases

- Add a condition for filtering
- Press X twice to delete condition
- Navigate to "Others" tab
- Navigate back - All Changes table doesn't crash.

Closes #4013


src/frontend/src/pages/ChangeRequestsPage/ChangeRequestsTable.tsx
```diff
@@ -177,10 +177,14 @@ const ChangeRequestsTable: React.FC = () => {
...
-    // sets filter to a default value if no filter is stored in local storage
-    localStorage.getItem('changeRequestsTableFilter') ?? '{"columnField": "identifier", "operatorValue": "=", "value": ""}'
-  );
+  const defaultFilter = '{"columnField": "identifier", "operatorValue": "=", "value": ""}';
+  let filterValues: { columnField: string; operatorValue: string; value: string };
+  try {
+    filterValues = JSON.parse(localStorage.getItem('changeRequestsTableFilter') ?? defaultFilter);
+  } catch {
+    localStorage.removeItem('changeRequestsTableFilter');
+    filterValues = JSON.parse(defaultFilter);
```
Steph375: One edge case: if localStorage contains valid JSON but not the expected shape, this will still succeed and give us undefined values in initialState. Might be worth validating the parsed object before using it.

Steph375 (changes requested):
This looks good to me just one small change! Great job!

patriots1:
Added checks to enforce valid JSON structure

cielbellerose (dismissed):
looks good to me!

src/frontend/src/pages/ChangeRequestsPage/ChangeRequestsTable.tsx
```diff
@@ -177,10 +177,20 @@ const ChangeRequestsTable: React.FC = () => {
...
 
-  const filterValues = JSON.parse(
-    // sets filter to a default value if no filter is stored in local storage
-    localStorage.getItem('changeRequestsTableFilter') ?? '{"columnField": "identifier", "operatorValue": "=", "value": ""}'
-  );
+  const defaultFilter = { columnField: 'identifier', operatorValue: '=', value: '' };
+  let filterValues: { columnField: string; operatorValue: string; value: string };
+  try {
+    const parsed = JSON.parse(localStorage.getItem('changeRequestsTableFilter') ?? 'null');
+    if (parsed && typeof parsed.columnField === 'string' && typeof parsed.operatorValue === 'string') {
```
wavehassman: should also check if value is a string

src/frontend/src/pages/GanttPage/ProjectGanttChart/GanttChartFilters.tsx
```diff
@@ -179,5 +179,4 @@ const GanttChartFilters = ({
     </Box>
   );
 };
```
wavehassman, line 180: also add this line back, not what you were working on, but needed

wavehassman (changes requested):

patriots1:
adressed previous comments

wavehassman (approved):
lgtm!

PR #4117, opened 2026-04-02
Fixed toast error with sabo number!

### LINKED ISSUE #4026: [Maintenance] - Add Toast Error for Sabo # Unique Contraint
## Description
In the Reimbursement Requests detail page, clicking the "Add SABO #" action button and submitting a SABO number that is already assigned to another reimbursement request causes a silent failure. The console says `Unique constraint failed on the fields: (`saboId`)` but the modal closes with no failure and the user has no way of knowing their sabo number wasn't updated

We want to catch this in the backend with a proper error message, and then display that error message on the frontend. This will inform the user when they try to submit that the saboid is already in use

## Proposed Solution
**Backend** _reimbursement-requests.services.ts_
- wrap the service method in a try catch and throw a specific exception when the unique constraint fails
- add an informative error message
- follow the patterns used in other service methods

**Frontend** AddSABONumberModal.tsx
- The toast error message should show up automatically when the backend throws the error since the onSubmit handler already calls toast.error
- ensure this works properly and add a screenshot of the message

## Acceptance Criteria
- [ ] Submitting a SABO number that already exists returns a proper Toast error
- [ ] Screenshot added to pr with Toast error
- [ ] Submitting a new unique SABO number still works as before

## Screenshot
Image

## Bug Log
```

→ 707 const reimbursementRequestWithSaboNumber = await prisma.reimbursement_Request.update(

Unique constraint failed on the fields: (`saboId`)

    at ei.handleRequestError (/node_modules/@prisma/client/src/runtime/RequestHandler.ts:228:13)

    at ei.handleAndLogRequestError (/node_modules/@prisma/client/src/runtime/RequestHandler.ts:174:12)

    at ei.request (/node_modules/@prisma/client/src/runtime/RequestHandler.ts:143:12)

    at async Array.$allOperations (/src/backend/src/prisma/prisma.ts:24:22)

    at async a (/node_modules/@prisma/client/src/runtime/getPrismaClient.ts:833:24)

    at async Function.setSaboNumber (/src/backend/src/services/reimbursement-requests.services.ts:707:48)

    at async setSaboNumber (/src/backend/src/controllers/reimbursement-requests.controllers.ts:235:7)

prisma:error 

Invalid `prisma.reimbursement_Request.update()` invocation in

/src/backend/src/services/reimbursement-requests.services.ts:707:83
```

nigelpurvis:
## Changes

On the frontend, I changed AddSABONumberModal.tsx: 
- Put onHide() inside the try so that the modal stays open if there's an error instead of closing silently.

Also on the frontend, in finance.api.ts:
- I added an await to the axios.post call in setSaboNumber so errors actually propagate instead of silently going through.

In the backend, in reimbursement-requests.services.ts:
- Wrapped the prisma.reimbursement_Request.update in a try,catch that detects the P2002 error and throws an HttpException instead of a raw Prisma error.

Now when theres a duplicate SABO number, the user sees an error bar on their screen explaining that they can't assign that number since it's already assigned.  
                                                                                                                                                                                         

## Screenshots
Screenshot 2026-04-02 at 3 07 31 PM

## Checklist

- [ ] No merge conflicts
- [ ] All checks passing

Closes #4026


cielbellerose (dismissed):
Great job!! small misspelling but went ahead and fixed

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -699,14 +699,20 @@ export default class ReimbursementRequestService {
     if (reimbursementRequest.organizationId !== organization.organizationId)
       throw new InvalidOrganizationException('Reimbursement Request');
 
-    const reimbursementRequestWithSaboNumber = await prisma.reimbursement_Request.update({
-      where: { reimbursementRequestId },
-      data: {
-        saboId: saboNumber
+    try {
```
chpy04: We should never be trying to catch in the service layer; that should always happen in the controller layer. Also, in this case, instead of using the prism type to catch the specific prism type error for when there's a collision on the unique key, we should just add a check before we go to update to see if anything else exists and then explicitly throw an error based on that.

chpy04 (changes requested):

src/backend/src/services/reimbursement-requests.services.ts
nigelpurvis: Fixed it! Now it just checks if the specified sabo number already exists, and if it does, throw HttpException, which the controller catches.

devContainerization/Dockerfile.backend.dev
```diff
@@ -14,4 +14,4 @@ COPY ./src/shared src/shared
 RUN yarn prisma:generate
 
 EXPOSE 3001
-CMD [ "sh", "-c", "yarn workspace backend dev" ]
\ No newline at end of file
+CMD [ "sh", "-c", "yarn workspace backend dev" ]
```
wavehassman: this shouldn't be changed

src/frontend/src/apis/finance.api.ts
```diff
@@ -363,7 +363,7 @@ export const getPendingAdvisorList = () => {
  * @param saboNumber the SABO number to set
  */
 export const setSaboNumber = async (requestId: string, saboNumber: string) => {
-  axios.post(apiUrls.financeSetSaboNumber(requestId), {
+  await axios.post(apiUrls.financeSetSaboNumber(requestId), {
```
wavehassman: should be return not await
both return and await will propagate errors, but return is more consistent with the other functions

src/backend/src/services/reimbursement-requests.services.ts
```diff
@@ -699,13 +699,19 @@ export default class ReimbursementRequestService {
     if (reimbursementRequest.organizationId !== organization.organizationId)
       throw new InvalidOrganizationException('Reimbursement Request');
 
+    const existingWithSaboNumber = await prisma.reimbursement_Request.findFirst({
```
wavehassman, line 702: add backend tests pls

wavehassman (changes requested):
Awesome job! Just a couple of changes

src/backend/src/services/reimbursement-requests.services.ts
nigelpurvis: Fixed those, and added some tests!

chpy04 (approved):
lgtm

PR #4128, opened 2026-04-02
#3995 blank design review title

### LINKED ISSUE #3995: [Maintenance] - Bug Design Review w/o Name
## Description
Calendar events require a title but the title currently can be " ", we should trim before checking if the title is empty

A logic addition like this should be in the backend, but we can look at _frontend/src/pages/CalendarPage/Components/EventModal.tsx_ to get a better understanding of the frontend issue here.
In this modal, currently we can enter " " as the title and it submits successfully
Image
We want to show a Toast error when a user tries to create/edit a calendar event with an empty title

## Proposed Solution
Ensure that the title **trimmed** is empty when creating/editing calendar events

Recommended to implement this logic in the backend - look at _calendar.routes.ts_

**Optional step**: Update the formatting on the toast error message to clearly indicate to the user that the title cannot be empty
## Acceptance Criteria 
- [ ] we cannot create an event with " " as a title, error message appears
- [ ] we cannot edit an event with " " as a title, error message appears
- [ ] title names such as "Finishline Design Review" still include spaces between words

## Bug Screenshot
Image

natashajoshi23:
## Changes

Ensured that design review title will not be blank by checking after trimming that length is greater than 1

## Screenshots

Screenshot 2026-04-02 at 7 19 52 PM

Closes #3995 (issue #)

### DIFF
--- a/src/backend/src/services/calendar.services.ts
+++ b/src/backend/src/services/calendar.services.ts
@@ -271,8 +271,6 @@ export default class CalendarService {
     zoomLink?: string,
     description?: string
   ): Promise<Event> {
-    if (!title.trim()) throw new HttpException(400, 'Title cannot be only whitespace');
-
     // Validate eventTypeId
     const foundEventType = await prisma.event_Type.findUnique({
       where: { eventTypeId }
@@ -594,8 +592,6 @@ export default class CalendarService {
     zoomLink?: string,
     description?: string
   ): Promise<Event> {
-    if (!title.trim()) throw new HttpException(400, 'Title cannot be only whitespace');
-
     // validate eventId
     const foundEvent = await prisma.event.findUnique({
       where: { eventId },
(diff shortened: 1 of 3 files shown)


wavehassman (changes requested):
We should also check this in the backend

src/backend/src/services/calendar.services.ts
```diff
@@ -271,6 +271,8 @@ export default class CalendarService {
     zoomLink?: string,
     description?: string
   ): Promise<Event> {
+    if (!title.trim()) throw new HttpException(400, 'Title cannot be only whitespace');
```
wavehassman: I would prefer to add this logic in validation.utils, so anytime we use nonEmptyString it checks for this

wavehassman (changes requested):

wavehassman (approved):
nice job!

PR #4140, opened 2026-04-08
#3929-Gantt-Chart-Improvements fix WP showing multiple times if it has multiple dependencies 

### LINKED ISSUE #3929: [Gantt Chart Improvements] - Duplicate Work Package Dependencies
### Description

If a work package depends on two other work packages, it should not show up twice

### Acceptance Criteria

A work package no longer shows up twice if it depends on two other packages

### Proposed Solution

Modify the frontend to change the display logic to the correct design

### Mocks

N/A

RemainingDelta:
## Changes
Previously, a Work Package (WP) with multiple dependencies would render once per dependency, causing the same WP bar to appear multiple times on the Gantt chart. This PR fixes that by ensuring each WP renders exactly once, with red arrows drawn from all its dependencies to that single instance.

## Notes
- Uses `react-archer` for arrow rendering
- WBS numbers determine which parent a blocked WP renders under

## Test Cases
- A WP with a single dependency shows one arrow
- A WP with multiple dependencies renders once with multiple arrows pointing to it
- A WP with no dependencies renders normally with no arrows

## Screenshots
image
image

Closes #3929


src/frontend/src/hooks/work-packages.hooks.ts
```diff
@@ -36,7 +36,17 @@ export const useAllWorkPackages = (queryParams?: { [field: string]: string }) =>
 export const useAllWorkPackagesPreview = (status?: string) => {
   return useQuery<WorkPackagePreview[], Error>(['work packages', 'preview', status], async () => {
     const { data } = await getAllWorkPackagesPreview(status);
```
chpy04, line 38: As far as I can tell there is no reason this endpoint should be returning duplicate work packages so we should be able to remove the changes on this file. If you have reason to believe otherwise there is a larger issue

chpy04 (changes requested):

src/frontend/src/hooks/work-packages.hooks.ts
RemainingDelta: Oh whoops. This was the first fix I implemented to fix this issue (it didn't work). I forgot to change it back. Let me revert it and then commit.

PR #4164, opened 2026-04-17
added check to enforce start date

### LINKED ISSUE #3934: [Maintenance] - Enforce Task Start Date <= Due Date
### Description

Currently task due dates can be set before their start date. Enforce that if a start date exists, it must be before or the same day as the due date. Ensure this works for both creating tasks and editing tasks.

Note: the logic for this is recommended to be handled in the backend

### Acceptance Criteria

- [ ] Toast error if start date is after the due date for creating and editing tasks

### Proposed Solution
The frontend modal can be found at `ProjectDetailPage/ProjectViewContainer/TaskList/TaskFormModal.tsx` 
The backend can be found in the files prefixed with _tasks_ for route/controller/service

### Mocks

Image

HLaPides:
## Changes

Added an if statement that raises an error if the start date is after the deadline. The if statement also checks if there is a start or end date so it doesn't raise an error if you create a project without a start or end.

## Notes

I'm still not sure if I did postman correct, I tried my best and it seems like it works

## Test Cases

Case 1: Start date before deadline. No error raised
Case 2: Start date after deadline. Error raised
Case 3: Start date on the deadline. No error raised

Closes #3934


src/frontend/src/pages/CalendarPage/Components/CalendarCreateTaskModal.tsx
```diff
@@ -21,7 +21,14 @@ const schema = yup.object().shape({
   status: yup.mixed<TaskStatus>().oneOf(Object.values(TaskStatus)).required(),
   assignees: yup.array().of(yup.string().required()).required(),
   startDate: yup.date().optional(),
-  deadline: yup.date().optional(),
+  deadline: yup
+    .date()
+    .optional()
+    .test('deadline-after-start', 'Deadline must be on or after the start date', function (deadline) {
```
wavehassman, line 27: instead of doing this, lets make the date picker not show dates before the start date

wavehassman (changes requested):
Great job!

wavehassman (approved):
very nice! used this pr to also make some bug fixes

PR #4224, opened 2026-05-20
4218-uploads-not-displaying-on-parts-review-page clear stale files state when part form modals close

### LINKED ISSUE #4218: [Maintenance] - Uploads not displaying on Parts Review page
### Description

On the parts review page when I upload something it doesn't show up. 

### Acceptance Criteria

- all uploads on the Parts Review page are successfully displayed after submission

### Proposed Solution

- Look at the code in PartsReviewPage.tsx
- Reproduce the error and identify where in the process its failing
- Implement a fix and verify uploads display correctly

### Mocks

_No response_

RemainingDelta:
## Changes
Fixed a bug in `SubmissionFormModal` and `ReviewFormModal` where uploaded file names would not appear after form submission. `NERFormModal` resets the RHF `fileIds`field array on every close, but the parallel local `files: File[]` state was never cleared. On re-open, the length mismatch caused the file list to not render. Added a `useEffect` in both modals that resets `files` (and `selectedPartIndex` in `ReviewFormModal`) whenever `open` becomes `false`. 

## Test Cases
- Repeat several times to ensure the fix works
  - Upload files in New Submission modal (`Projects` -> Specific Project -> `Part Review` -> `New` -> `New Submission`) — file names appear in the list                                                                                                   
  - Submit the form, re-open New Submission — file list is empty with no stale entries                                                                                     
  - Upload a new file on re-open — correct file name appears                                                                                                               
  
### Before
**Note: This sometimes happens, not always**
A file is uploaded, but it doesn't show file name
New Submission
All Parts for Impact Attessatel

### After
A file is uploaded, but it **always** shows file name:
New Submission
All Parts for legact Altengate

Closes #4218

### DIFF
--- a/src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/PartReview/PartReviewComponents/PartFormModels/ReviewFormModal.tsx
+++ b/src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/PartReview/PartReviewComponents/PartFormModels/ReviewFormModal.tsx
@@ -3,7 +3,7 @@ import { useToast } from '../../../../../../hooks/toasts.hooks';
 import * as yup from 'yup';
 import { Controller, useFieldArray, useForm } from 'react-hook-form';
 import { yupResolver } from '@hookform/resolvers/yup';
-import React, { useEffect, useState } from 'react';
+import React, { useState } from 'react';
 import NERFormModal from '../../../../../../components/NERFormModal';
 import { Autocomplete, Button, Grid, IconButton, List, ListItem, Typography } from '@mui/material';
 import { FormControl, FormHelperText, FormLabel, TextField } from '@mui/material';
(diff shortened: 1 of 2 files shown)


staysgt (dismissed):
Good job!

src/frontend/src/pages/ProjectDetailPage/ProjectViewContainer/PartReview/PartReviewComponents/PartFormModels/ReviewFormModal.tsx
```diff
@@ -25,6 +25,13 @@ const ReviewFormModal = ({ open, handleClose, defaultValues, onSubmit, partsInPr
   const [uploading, setUploading] = useState(false);
   const { mutateAsync: uploadFile } = useUploadFile();
 
+  useEffect(() => {
```
wavehassman: Instead of a useEffect in both files, just wrap handleClose in a callback that also clears the state. Then update the ```onHide=``` call. It should look something like this: 
```
const handleCloseAndReset = () => {
  setFiles([]);
  setSelectedPartIndex(undefined); 
  handleClose();
};

wavehassman (changes requested):
Great job fixing the bug! I would just prefer it to be structured differently

wavehassman (approved):
perfect!

PR #4231, opened 2026-05-23
upgraded yarn to add release age min and added min to npm

wavehassman:
## Changes

Upgrades Yarn from 3.2.1 to 4.15.0 with a 3-day minimum package release age enforced in both Yarn and npm as a supply chain security measure, updates Dockerfiles for compatibility, and migrates Prisma config from package.json to prisma.config.ts to resolve a deprecation warning ahead of Prisma 7.

Slack messages still work and uploading documents

## Checklist

- [x] No merge conflicts
- [x] All checks passing


src/backend/Dockerfile
```diff
@@ -1,16 +1,16 @@
 # TO BE RUN FROM DOCKER COMPOSE. DO NOT RUN MANUALLY AS CONTEXT IS NOT SET CORRECTLY
 FROM platformatic/node-caged:25-slim
-RUN npm install -g yarn
```
chpy04, line 3: The reason that we install yarn before copying everything over is because of docker caching, so we ideally want the yarn installation to be as early as possible to leverage caching as much as possible, specifically before copying backend and shared bc those change a lot

src/backend/Dockerfile
```diff
@@ -1,16 +1,16 @@
...
 COPY ./src/backend src/backend
 COPY ./src/shared src/shared
 
 RUN rm -rf src/backend/tests
 
-RUN yarn install --prod;
+RUN npm install -g corepack --force && corepack enable && yarn install
 
 RUN echo DATABASE_URL=\"postgresql://postgres:docker@database:5432/nerpm?schema=public\" > ./src/backend/.env
 RUN echo NODE_ENV=development >> ./src/backend/.env
```
chpy04, line 18: If you are changing this dockerfile you will almost certainly need to change the development one to keep the development env working as well

src/frontend/Dockerfile
```diff
@@ -1,15 +1,16 @@
 # TO BE RUN FROM DOCKER COMPOSE. DO NOT RUN MANUALLY AS CONTEXT IS NOT SET CORRECTLY
-FROM node:20 AS builder
+FROM node:22 AS builder
```
chpy04: we really ought to be on the same version of node for the frontend and backend, although I must have missed this when Raphael added the caged node 25. Bc you are already doing dependency work, can you upgrade one to 25 or downgrade one to 22 (idc which ideally 25 but if you run into problems 22 is fine)

.dockerignore
```diff
@@ -61,8 +62,6 @@ yarn-error.log*
 
 # Lock files (keep package.json though)
 package-lock.json
-yarn.lock
-.yarnrc.yml
```
chpy04, line 65: Why do we need these? I would have thought the releases folder would contain everything we need

chpy04 (changes requested):
We technically are fully es modules, although the only reason we switched from common js on the frontend was bc my deployment pipeline broke and I couldn't figure out what the issue was so I did that instead so maybe its fine. If the cjs works fine its good w me but other comments should be addressed

.dockerignore
wavehassman: .yarnrc.yml points yarn at the release binary (new with yarn 4), and yarn.lock pins dependency versions for reproducible builds

chpy04 (approved):

PR #4304, opened 2026-07-04
#4215 Add work package factory and process

### LINKED ISSUE #4215: [Seed Data] Generate Work Packages with Timeline Logic
### Description

- Generate work packages for projects using the project timelines from seed context.
- Work packages should create realistic project structure and support later task generation.
- Include functionality where earlier work packages can block later work packages.
- Each work package should be created with its required WBS element.

### Acceptance Criteria

- Each project gets 0–8 work packages.
- Average work package count is around 5.
- Work packages belong to valid projects via `projectId`.
- Each work package has a valid `wbsElementId`.
- Each work package has a valid `orderInProject`.
- Work package dates stay inside the project timeline.
- Blocking relationships only reference earlier work packages in the same project.
- Blocked work packages start after blocker end dates.
- Generated work packages are stored in `ctx.workPackages`.

### Proposed Solution

- Add `prisma/seed/factories/workPackageFactory.ts`
- Create `createSeedWorkPackage(ctx, project, overrides?)`
- Create `generateSeedWorkPackages(ctx)`
- Use `ctx.projectTimelines[project.projectId]` to determine valid date ranges.
- Create a `WBS_Element` before creating each `Work_Package`.
- Store generated work package timelines in context so task generation can use them later.

Santiordon:
## Changes

Added factory and process for work package generation on top of projects.

## Checklist

- [x] No merge conflicts
- [ ] All checks passing

Closes #4215 (issue #4215)


Santiordon:
Dependency added: ora (an elegant terminal spinner for a cleaner and modern look for the seed generation).

src/backend/src/prisma/factories/work-package.factory.ts
```diff
@@ -0,0 +1,84 @@
+import { Faker } from '@faker-js/faker';
+import { Prisma, WBS_Element_Status, Work_Package_Stage } from '@prisma/client';
+import { DateRange } from '../context.js';
+import { clampDate, daysBetween } from '../dates.js';
+import { addDaysToDate } from 'shared';
+
+const DAYS_PER_WEEK = 7;
```
Steph375: Cant we just use the one from dates

package.json
```diff
@@ -77,6 +77,7 @@
     "@types/multer": "^1.4.12",
     "canvas-confetti": "^1.9.3",
     "mitt": "^3.0.1",
+    "ora": "^9.4.1",
```
Steph375: since ora is imported from backend seed code, should it be declared in the backend workspace package rather than only in the root package.json not sure just a thought

src/backend/src/prisma/factories/work-package.factory.ts
```diff
@@ -0,0 +1,84 @@
...
+      : addDaysToDate(
+          projectTimeline.start,
+          faker.number.int({
+            min: 0,
+            max: Math.max(0, daysBetween(projectTimeline) - DAYS_PER_WEEK)
+          })
+        );
+
+  // duration saved in WEEKS instead of days
+  const maxDuration = Math.max(1, daysBetween({ start, end: projectTimeline.end }) / DAYS_PER_WEEK);
```
Steph375: maxDuration can be fractional because it divides days by DAYS_PER_WEEK, but then it is passed to faker.number.int. Could we floor it first so the max is definitely an integer?

src/backend/src/prisma/factories/work-package.factory.ts
```diff
@@ -0,0 +1,84 @@
...
+    { weight: 9, value: 7 },
+    { weight: 5, value: 8 }
+  ]);
+
+export const generateWorkPackageStage = (faker: Faker): Work_Package_Stage =>
+  faker.helpers.arrayElement(Object.values(Work_Package_Stage));
+
+export const generateWorkPackageTimeline = (faker: Faker, projectTimeline: DateRange, blockerEndDate?: Date): DateRange => {
+  const start =
+    blockerEndDate && blockerEndDate < projectTimeline.end
```
Steph375, line 117: Could there be an edge case here when `blockerEndDate` is at or after `projectTimeline.end`? Since that would fall back to a random start date, it seems possible to create a blocked work package that starts before the blocker ends. Should we avoid creating a blocked work package in that case, or clamp/start it differently?

Steph375 (changes requested):
Looks good just a few things! Also not sure if I'm missing something but I see the project template configs include workPackageTemplates, but I don’t see those being created/connected in projectTemplateCreateInput. Are those work package templates created somewhere else that I missed, or should that be added there?

package.json
Santiordon: i assume its in this one? i just did yarn install and it put it there. might be good to get waverly's thoughts though.

src/backend/src/prisma/factories/work-package.factory.ts
Santiordon: Good call.

src/backend/src/prisma/factories/work-package.factory.ts
Santiordon: Yep. this is addressed in the process file:

"if the blocker ends too close to the project end, drop the blocking relationship. this would skew the data with 5-10% of blocking relationships being affected in the worst case scenario. this happening is extremely rare but does change the data slightly if it does occur."

yarn.lock
```diff
@@ -13891,6 +13907,7 @@ __metadata:
     eslint-plugin-cypress: "npm:4.3.0"
     eslint-plugin-prettier: "npm:^5.2.2"
     mitt: "npm:^3.0.1"
+    ora: "npm:^9.4.1"
```
wavehassman: yeah this should go in the backend

src/backend/src/prisma/processes/seed-runner.ts
```diff
@@ -20,35 +21,65 @@ export class SeedRunner {
...
+      } catch (e) {
+        const elapsed = `${((Date.now() - start) / 1000).toFixed(2)}s`;
+        spinner.fail(`${index} ${name}   ${elapsed}`);
+        failures++;
+        throw e;
+      }
+    }
 
-      console.log(`${instance.constructor.name} complete`);
+    const totalElapsed = `${((Date.now() - totalStart) / 1000).toFixed(2)}s`;
```
wavehassman, line 77: move this to the catch like 
```
} catch (e) {
  const elapsed = `${((Date.now() - start) / 1000).toFixed(2)}s`;
  spinner.fail(`${index} ${name}   ${elapsed}`);
  const totalElapsed = `${((Date.now() - totalStart) / 1000).toFixed(2)}s`;
  console.log();
  console.log(`  ❌ Seed failed on ${instance.constructor.name.trim()} after ${totalElapsed}`);
  console.log();
  throw e;
}
```

The failures counter is never hit because it will throw

wavehassman (changes requested):
I think there should be more templates and work packages at a null stage

yarn.lock
Santiordon: Okidokes

src/backend/src/prisma/processes/seed-runner.ts
Santiordon: Fixed

wavehassman (changes requested):
looking back at this, I don't love how they're all WP1..2...3... and so forth. they are usually like design, manufacturing, etc. also no work packages are blocked by others, idk if that is a later ticket, but we should definitely have some blockers. there are no upcoming work packages, they are all hella overdue. can we make the deadlines more realistic?

also for the current projects, all work packages are active. only one or two should be active at a time normally, and if something is blocking another thing that can't be activated until the first one is completed

Image

wavehassman (approved):
perfect!

Issue #4404, opened 2026-08-29
Athena MCP


chpy04:
MCP server for new athena application. Pr will be split into following changes:
1. api token creations + mcp router with read only auth guards & single health check route
2. Exposed read only routes in mcp for agent content
3. MCP wrapper on top of api to implement protocol for external usage

PR #4405, opened 2026-08-29
#4404 api tokens

chpy04:
## Changes

Added per user scoped api tokens that exclusively access an mcp router. Router will implement mcp protocol in later commits.  Tokens are hashed in backend and stored to user. Can be regenerated to  revoke previous token.

## Notes

Requires custom auth, tokens are completely separate from normal auth (token auth cannot touch regular endpoints and regular auth cannot touch mcp endpoints

## Screenshots
Screenshot 2026-08-29 at 2 38 51 PM


src/frontend/src/pages/SettingsPage/ApiKey/ApiKeySection.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const toast = useToast();
+
+  const [generatedToken, setGeneratedToken] = useState<string | undefined>();
+  const [showRegenerateConfirm, setShowRegenerateConfirm] = useState(false);
+
+  if (isError) return <ErrorPage message={error?.message} />;
+  if (isLoading) return <LoadingIndicator />;
+
+  const onGenerate = async () => {
+    setShowRegenerateConfirm(false);
```
wavehassman: move inside the try so if generateApiToken fails, the modal stays open

src/backend/src/prisma/migrations/20260829170504_add_user_api_tokens/migration.sql
```diff
@@ -0,0 +1,22 @@
+-- CreateTable
+CREATE TABLE "User_API_Token" (
```
wavehassman, line 2: I think you should put a unique index on (userId) WHERE dateRevoked IS NULL, so we make sure only one is live at a time

wavehassman (changes requested):

src/backend/src/prisma/migrations/20260829170504_add_user_api_tokens/migration.sql
chpy04: Why? Right now there is only one, but the whole point of the new table is that if we wanted a many to one relationship later we can, and the auth side already supports multiple so this seems like something that would just cause us an unecessary future migration if we did want multiple in the future

src/frontend/src/pages/SettingsPage/ApiKey/ApiKeySection.tsx
chpy04: ok

PR #4406, opened 2026-08-29
#4404 initial routes for the mcp

chpy04:
## Changes

routes for the mcp

### DIFF
--- a/src/backend/src/controllers/mcp.controllers.ts
+++ b/src/backend/src/controllers/mcp.controllers.ts
@@ -1,5 +1,7 @@
 import { NextFunction, Request, Response } from 'express';
 import { getRoleInOrganization } from '../utils/mcp-auth.utils.js';
+import { HttpException } from '../utils/errors.utils.js';
+import McpService from '../services/mcp.services.js';
 
 export default class McpController {
   /**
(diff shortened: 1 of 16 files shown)


src/backend/src/controllers/mcp.controllers.ts
```diff
@@ -29,4 +31,65 @@ export default class McpController {
       next(error);
     }
   }
+
+  static async getProjectsByCarNumber(req: Request, res: Response, next: NextFunction) {
+    try {
+      const { carNumber } = req.params as Record<string, string>;
+      const parsedCarNumber = Number(carNumber);
+
+      if (!Number.isInteger(parsedCarNumber) || parsedCarNumber < 0) {
```
wavehassman, line 40: should also make sure this 32-bit for it to be valid
might want to add that to validateWBS as well

src/backend/src/services/mcp.services.ts
```diff
@@ -0,0 +1,178 @@
...
+  }
+
+  /**
+   * Gets the events scheduled within a date range.
+   * @param startDate the start of the range
+   * @param endDate the end of the range
+   * @param organization the organization the request is scoped to
+   * @throws if the range is inverted or wider than a week
+   */
+  static async getEvents(startDate: Date, endDate: Date, organization: Organization): Promise<McpEvent[]> {
```
wavehassman, line 160: this will include events on the final day of the range right?

wavehassman (changes requested):

src/backend/src/controllers/mcp.controllers.ts
chpy04: 32 bit check is redundant, added wbs regex validation

src/backend/src/services/mcp.services.ts
chpy04: will now

PR #4408, opened 2026-08-29
full mcp implementation

### LINKED ISSUE #4404: Athena MCP

chpy04:
## Changes

The actual mcp implementation necessary for agents to connect

## Notes
- needed to add zod for mcp (we should only use this for mcp)
- Decided not to implement oauth2.0 (technically what mcp wants) and opt for api token approach for simplicity (still configurable in claude with:
```
claude mcp add --transport http finishline https://api.finishlinebyner.com/mcp \ 
  -H "x-api-key: fl_api_key"
```

## Screenshots

Screenshot 2026-08-29 at 6 13 56 PM

Closes #4404


src/backend/src/mcp/tools.ts
```diff
@@ -0,0 +1,131 @@
...
+        'The range cannot be wider than 7 days; ask for one week at a time.',
+      inputSchema: z.object({
+        startDate: z.string().describe('Start of the range, as an ISO date such as "2026-09-01".'),
+        endDate: z
+          .string()
+          .describe('End of the range, as an ISO date. Must be on or after startDate and no more than 7 days later.')
+      }),
+      annotations: { readOnlyHint: true }
+    },
+    async ({ startDate, endDate }) =>
```
wavehassman, line 187: could the model send something like "next Monday" Wouldn't an invalid date slip past both range checks since NaN comparisons are false, and then it would be a 500 that says 'report this to the software team'

src/backend/index.ts
```diff
@@ -94,7 +96,11 @@ app.use('/ics', icsRouter);
 // API token routes — mounted before the JWT middleware so that per-user API tokens authenticate here
 // and ONLY here. Keeping this above the cookie middleware is what stops a token from reaching the
 // rest of the API.
-app.use('/mcp', mcpRouter);
+app.use('/agent', agentRouter);
+
+// The MCP endpoint, authenticated with the same per-user API tokens. It must accept POST (JSON-RPC),
+// so it is mounted separately from the read only /agent router; every tool it exposes is a read.
+app.all('/mcp', requireApiToken, attachAuthInfo, (req, res) => void mcpNodeHandler(req, res, req.body));
```
wavehassman: maybe i'm confused but this doesn't seem to have a readOnly guard, only a requireApiToken guard. If someone later registers a tool that writes, could any API token reach it?

wavehassman: shouldn't this have an onerror handler?

src/backend/src/mcp/handler.ts
```diff
@@ -0,0 +1,19 @@
...
+ *
+ * The factory runs once per request, so each request gets a fresh server built for its own caller
+ * and nothing is retained between requests. That is what makes the endpoint stateless and safe to
+ * run behind a load balancer with more than one instance.
+ *
+ * responseMode 'json' pins plain JSON responses instead of upgrading to server sent events; we emit
+ * no progress or logging notifications, and it keeps the endpoint easy to test with curl.
+ */
+const mcpHandler = createMcpHandler(({ authInfo }) => buildMcpServer(authInfo?.extra as unknown as AgentContext), {
+  responseMode: 'json'
```
wavehassman, line 38: shouldn't we check that authInfo is there and give a good error if not

src/backend/src/mcp/errors.ts
```diff
@@ -0,0 +1,52 @@
...
+import { HttpException } from '../utils/errors.utils.js';
+
+/** Cap on array results, so one oversized project cannot flood the model's context window. */
+const MAX_ITEMS = 100;
+
+/**
+ * Serializes a tool result, truncating long lists.
+ * @param value the value to return to the model
+ */
+export const toolJson = (value: unknown): CallToolResult => {
```
wavehassman: is the MAX_ITEMS cap still doing anything here? looks like it only truncates when the value is an array, but getProjects returns { carNumber, projects } now, so finishline_list_projects might not get capped at all

wavehassman (changes requested):
Since the caller here is a model rather than our UI, the inputs aren't really controlled. We could get 01.1.0, 1.1, sometimes a WBS that doesn't exist. Right now wbsNum is an unconstrained z.string(). I think adding a regex to the tool schema would help return a proper validation error
```
z.string().regex(/^\d+\.\d+\.\d+$/, 'WBS must look like 1.1.0') 
```
This would fix the validation issues I flagged in the other pr

src/backend/src/mcp/tools.ts
chpy04: done

src/backend/index.ts
chpy04: The guard is not on the auth, its just a byproduct of the fact that api token auth can only access the /mcp and /agent routes, which only have readonly calls. Eventually we will want to add write abilities, and so this scoping control is the best way to separate

src/backend/src/mcp/handler.ts
chpy04: This is not user input so that wouldn't really be helpful

src/backend/src/mcp/errors.ts
chpy04: changed to pagination

src/backend/index.ts
chpy04: done

PR #4523, opened 2026-09-17
#4511 competition performance

### LINKED ISSUE #4511: [Executive Summary] - Competition Performance section
### Description

We want to create a competition performance section. This component should exist in its own file. This should display final place, total points earned, best/worst static event, best/worst dynamic event, acceleration top time, endurance laps completed, autocross top time, endurance average lap time. It should separate these values by competition. The component should take in CompetitionPerformance and display the information as shown in the mock.

### Acceptance Criteria

- FSAE / FHE subtabs
- renders final place, total points earned, best/worst static event, best/worst dynamic event, acceleration top time, endurance laps completed, autocross top time, endurance average lap time

### Proposed Solution

- create new file to store component
- have component take in CompetitionPerformance from shared types
- display information according to the mock

### Mocks

Image

HLaPides:
## Changes

Created a competition performance component that meets the following criteria:

FSAE / FHE subtabs
renders final place, total points earned, best/worst static event, best/worst dynamic event, acceleration top time, endurance laps completed, autocross top time, endurance average lap time

## Notes

The Width was hardcoded at 500px to emulate the size of the component on the full executive summary mock, This is noted in comments and will be easy to change if/when it needs to be.

I added a check to make sure that both endurance laps and max laps existed before rendering the text. Idea was that if i just checked for endurance laps it could result in something like 40/N/A which would've caused the same readability issue.

## Test Cases

Tested to make sure it rendered placements properly e.g (1st, 2nd, 3rd, 4th, 11th, 12th, 21st)
Tested case where donut graph had no points to display
Tested case where endurance laps/ max laps were undefined.

## Screenshots

**Normal window size (FSAE):**
Fullscreen_competition

**Smallest window size:**
image

**FSAE tab:**
FSAE_Competition_Demo

**FHE tab:**
FHE_Competition_Demo

**No wheel or endurance laps:**
image

## To Do
1. For the visualization wheel I did not find a points per event in the schema so I just made it do the total number of points without a breakdown of points per event. Will need to be fixed when the schema is updated, but added the fix where it won't render if there is nothing to display.

Closes #4511


src/frontend/src/components/CompetitionPerformance.tsx
```diff
@@ -0,0 +1,194 @@
...
+                <UnderlinedLabel label="Total Points Earned" value={`${performance.totalPointsEarned ?? 'N/A'}`} />
+              </Box>
+
+              <UnderlinedLabel label="Best Static Event" value={performance.bestStaticEvent ?? 'N/A'} />
+              <UnderlinedLabel label="Worst Static Event" value={performance.worstStaticEvent ?? 'N/A'} />
+              <UnderlinedLabel label="Best Dynamic Event" value={performance.bestDynamicEvent ?? 'N/A'} />
+              <UnderlinedLabel label="Worst Dynamic Event" value={performance.worstDynamicEvent ?? 'N/A'} />
+            </Box>
+
+            <Box flex="1 1 140px" display="flex" justifyContent="center" alignItems="flex-start">
```
staysgt: Instead of having total points set to zero if it doesn't exist, I think we should instead only display the donut graph if there are points to display. Same with the max points; only display the donut graph if totalPointsEarned and maxPoints are both defined

src/frontend/src/components/CompetitionPerformance.tsx
```diff
@@ -0,0 +1,194 @@
...
+              <PointsRing earned={performance.totalPointsEarned ?? 0} max={PLACEHOLDER_MAX_POINTS} />
+            </Box>
+
+            <Box display="grid" gridTemplateColumns="1fr 1fr" gap={3} width="100%">
+              <StatBlock value={formatTime(performance.accelerationTopTimeSeconds)} label="Acceleration Top Time" />
+              <StatBlock
+                value={`${performance.enduranceLapsCompleted ?? 'N/A'}/${PLACEHOLDER_MAX_ENDURANCE_LAPS}`}
+                label="Endurance Laps Completed"
+              />
```
staysgt: For the endurance laps completed, if endurance laps is undefined and max laps is set, it will display something like: N/A/40, which is difficult to read. Instead we should only display this section if enduranceLapsCompleted is defined

staysgt (changes requested):
Looks really good so far!

staysgt (approved):
Really good work!

wavehassman (approved):
great job!

